PR #2, opened 2021-11-05
Add support for running estimator on binder

malb:


malb:
To enable testing the required Juypter notebooks we also switch to `pytest`.

PR #8, opened 2021-12-15
Dual attack and (better) support for hybrid attacks

blowfish880:
This adds estimations for the following algorithms:

### The dual attack 
Viewed as dimension-noise trade-off as in [Alb17](https://link.springer.com/chapter/10.1007/978-3-319-56614-6_4). The new noise is computed as in [EspJouKha20](https://link.springer.com/chapter/10.1007/978-3-030-65277-7_20).

### Brute-force algorithms
This includes exhaustive search and meet-in-the-middle attacks, mostly to combine with the dual attack to get hybrid attacks. For exhaustive search, the optimal number of samples is computed as in [EspJouKha20](https://link.springer.com/chapter/10.1007/978-3-030-65277-7_20). For the MITM attack we mostly follow [CHHS19](https://ieeexplore.ieee.org/abstract/document/8747481). 

### Primal Hybrid: Support Babai with accurate success probability
This adds support for enforcing Babai in the primal hybrid with an accurate estimate for the mitm probability from [SC19](https://dl.acm.org/doi/abs/10.1145/3338469.3358941).

### Misc
This also adds a convenience function `estimate_lwe` similar to the old estimator to run (a subset of) all attacks to estimate the security of a specific instance quickly.

The dual hybrids in this version do not yet include [QT21](https://link.springer.com/chapter/10.1007%2F978-3-030-92068-5_2), since it only just became available.


estimator/util.py
```diff
@@ -230,3 +275,7 @@ def batch_estimate(params, algorithm, jobs=1, log_level=0, **kwds):
         ret[x][f] = res[f, x]
 
     return ret
+
+
+def log2(x):
```
malb, line 280: Do we need this?

estimator/util.py
```diff
@@ -181,6 +181,51 @@ def binary_search(
         return it.y
 
 
+def robust_bin_search(
```
malb, line 184: Shall we absorb this into `local_minimum`?

estimator/top_level.py
```diff
@@ -0,0 +1,73 @@
+from .conf import red_cost_model_default, red_shape_model_default
+from .lwe_dual import dual, dual_hybrid, dual_mitm_hybrid
+from .lwe_primal import primal_usvp, primal_hybrid, primal_bdd
+from .gb import arora_gb
+from .lwe_brute_force import exhaustive_search, mitm
+
+
+def estimate_lwe(params,
```
malb, line 8: I've got an unpushed version of this, will merge the two

docs/schemes/hes.rst
```diff
@@ -17,5 +17,5 @@ Homomorphic Encryption Standard
    >>> HESv111024128ternary
    LWEParameters(n=1024, q=134217728, Xs=D(σ=0.82), Xe=D(σ=3.00), m=1024, tag='HESv11ternary')
    >>> primal_hybrid(HESv111024128ternary)
-   rop: ≈2^130.8, red: ≈2^130.5, svp: ≈2^128.4, β: 338, η: 369, ζ: 0, |S|: 1, d: 2049, prob: 1, ↻: 1, tag: hybrid
+   rop: ≈2^183.4, red: ≈2^182.6, svp: ≈2^182.3, β: 345, η: 2, ζ: 139, |S|: ≈2^220.3, ...
```
malb, line 20: So hybrid now enforces Babai? Shouldn't it allow large \eta, too?

estimator/util.py
blowfish880: It's heavily used in the MITM attack. Got tired of writing log(x, 2), which looks kinda ugly when x is some longer expression. I thought it might be useful more generally, but if you think it's superfluous, I can also remove it and either place it in lwe_brute_force.py or replace it with log(x, 2) everywhere.

estimator/util.py
blowfish880: I found `local_minimum` already somewhat complicated to understand, so including another layer of complexity might be bad for readability. We could replace `binary_search` by this though, since the default setting `step=1` is essentially the same thing.

estimator/util.py
malb: It's purely aesthetics on my part. I'd perhaps define it inside MITM? But it's no big deal either way.

estimator/util.py
malb: I personally find `with local_minium() as it: blah` much easier to read so I might try to turn your function into an iterator and then have `with local_minimum_simple` and `with local_minimum`?

blowfish880: I agree that `with local_minimum as it` is nice and easy to read, I just find the implementation of the class `local_minimum` not easy to follow and worry that including the step in there might make it even trickier. Especially, since it is very easy to implement this feature as a wrapper (as is done in `robust_bin_search`) and this provides a nice layer of abstraction. But maybe writing a new context class that uses the current one might be even more elegant.

estimator/top_level.py
bencrts: Feel free to completely overwrite -- this was a very quick first effort.

docs/schemes/hes.rst
bencrts: We decided to set `babai = True` to be the default purely because `prob.mitm_babai_probability()` only covers the case of `η = 2`. In the case of `η > 2`, this probability is yet-to-be analysed and so we've (conservatively) set this probability to be `1`. We thought about maybe setting `babai = False` and allowing the computation of `prob.mitm_babai_probability()` only when `η = 2`, but this might be hiding a few too many assumptions under-the-hood, or could potentially lead to weird optimizations.

Definitely open to alternatives!

estimator/lwe_dual.py
```diff
@@ -0,0 +1,450 @@
...
+        - ``red``: Number of word operations in lattice reduction.
+        - ``δ``: Root-Hermite factor targeted by lattice reduction.
+        - ``β``: BKZ block size.
+        - ``ζ``: Number of guessed coordinates.
+        - ``h1``: Number of non-zero components among guessed coordinates (if secret distribution is sparse)
+        - ``prob``: Probability of success in guessing.
+        - ``repetitions``: How often we are required to repeat the attack.
+        - ``d``: Lattice dimension.
+
+        - When ζ = 1 this function essentially estimates the dual attack.
```
malb, line 274: Why 1? Also `1 0`?

malb:
I've merged this, because it's obviously great! Thanks. I'll do some clean-up/follow-up on `main`.

PR #9, opened 2021-12-22
Use plain distinguisher for plain dual attack

blowfish880:
This adds a simple distinguisher to lwe_guess.py, which in turn is now used in the plain dual attack (zeta=0).


malb:
Ta!

PR #10, opened 2021-12-22
Fix typo (distinghuish -> distinguish)

joerowell:
I think I got all of them but it might be worth double checking.


malb:
Whoops!

blowfish880:
OMG! Can't believe I did that. Thanks, Joe!

PR #11, opened 2022-01-03
bug fix, DualHybrid: check for reduction in full_sieves was incorrect

blowfish880:
I think the current check is incorrect, since `DualHybrid.full_sieves` contains the names, not the functions for the relevant sieves.

But this points to the larger issue that the current way these different approaches of obtaining many short vectors from the reduction cost models are handled is somewhat hacky in the dual attack. Ideally, one would be able to request a certain number of short vectors from the model and the model would return the cost of obtaining all of them and what length they would have. Maybe the cleanest way would be to have cost models not being just functions, but objects with two methods: 1) something like `reduce_basis(beta)` that returns the cost of a reduction + the shape, and 2) something like `short_vectors(beta, N)` that returns the cost of obtaining N short vectors and their length. This would reflect the two different ways that BKZ is used in the different attacks: obtaining nice bases and as an approximate SVP oracle. But if the dual attack is the only place that comes up, not sure if it's worth it.


malb:
PR looks good to me, thanks! I'll open an issue to discuss the second point now, too.

PR #14, opened 2022-01-05
bug fix: dual(_hybrid) was failing on some NTRU instances

blowfish880:
The issue was that `dataclasses.replace` did not copy the hamming weight in ternary secrets (which only really was a problem when a SparseTernary distribution was not actually sparse). We already store the density and the dimension, and we have a function that derives the hamming weight from them, so let's simply use that instead.


blowfish880:
I also just added another small bug fix: the Kyber model freaks out when the block size is > 1024, which can happen during optimization for large `n` (see e.g. `schemes.CHHS_...`). But such a large block size is unlikely to be relevant so I think restricting it to be smaller than that should be fine.

malb:
I'm cool with picking a bound I'd go a bit more crazy, like 2048. Rationale: maybe people want 320 bits of security for some multi-target reason?

blowfish880:
In the first step of the loop `beta_upper` is doubled, so I guess block sizes between 1024 and 2048 are being considered during optimization (unknowingly to my former self). There is a comment in the `reduction.py` though:

```
# These are not asymptotic expressions but compress the data in [AC:AGPS20]_ which covers up to
# β = 1024
```
If I interpret this right, if values > 1024 are returned, they might be off. So should we print a warning in that case?

malb:
Ah, cool. That comment is mostly there to avoid overinterpretation of the data, i.e. some people might consider this a true asymptotic expression while it's "just" compressing a table. The interpolation will be reasonable in the neighbourhood of 1024, so I'd say using it up to 2048 is not ideal but okay. We could consider to compute this data for up to 2048 at some point. For now, I'd say let's merge.

malb:
Oh, can we add a test of something that failed and now works?

PR #17, opened 2022-01-09
Revise number of samples in Frodo parameters

dstebila:
Checked with my Frodo colleagues and I was overestimating the number of samples for the Frodo parameters. These are the correct numbers.


malb:
Ta.

PR #18, opened 2022-01-12
feat(lwe): add decoding variants to top level fn

bencrts:
Update the top level function so that it outputs four decoding variants:

- `bdd`, as before
- `primal_hybrid` which calls `primal_hybrid` with `mitm = False` and `babai = False`
- `primal_mitm_hybrid` which calls `primal_hybrid` with `mitm = True` and `babai = True`
- `gv_mitm_hybrid` which calls `primal_hybrid` with `mitm = True` and `babai = False`

 We can ignore the fourth variant with `mitm = False` and `babai = True`, since this case should be covered by `primal_hybrid`.

Thoughts on the namings? We could instead do something like `gv_mitm_hybrid` -> `bdd_mitm_hybrid`.

Keeping as a draft PR for now, to check the tests etc.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,9 +29,11 @@ Quick Start
 
     >>> r = LWE.estimate(Kyber512)
     arora-gb             :: rop: ≈2^inf, dreg: 25, mem: ≈2^106.3, t: 3, m: ≈2^inf, tag: arora-gb, ↻: ≈2^inf, ζ: 480
-    bkw                  :: rop: ≈2^178.8, m: ≈2^166.8, mem: ≈2^167.8, b: 14, t1: 0, t2: 16, ℓ: 13, #cod: 448, #top: 0, ...
+    bkw                  :: rop: ≈2^178.8, m: ≈2^166.8, mem: ≈2^167.8, b: 14, t1: 0, t2: 16, ℓ: 13, #cod: 448, #top: 0, #test: 64, tag: coded-bkw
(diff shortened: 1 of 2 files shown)


malb:
Should it be `bdd_hybrid` and `bdd_hybrid_mitm`? Why `gv_`?

bencrts:
Agreed about using `bdd_` over `primal_`.

The question then becomes how to differentiate the `babai=False` and `babai=True` cases when `mitm = True` (since they are not related in the same way as the `mitm=False` case). 

We could have:

`bdd_hybrid`
`bdd_mitm_hybrid`
`bdd_mitm_hybrid_babai`

(?)

Of course, if we can figure out the mitm proability (currently set to 1) for the `mitm=True, babai=False` case, then this problem goes away and we can merge the two techniques (as we have done for `mitm=False`). 

A second option would be to just ignore one of the `mitm=True` cases for now.

To be honest I don't really have a strong opinion on this, so I can update the PR to match whatever is preferable.

P.S: the `gv_` was just to match "guess and verify", although I agree that this is confusing here - and that we should stick to `bdd_`.

malb:
Yeah, maybe we ignore the overly optimistic variant for now and leave a comment in the code for why?

bencrts:
Should be OK now hopefully -- will try to remember to finish everything locally before pushing next time.

malb:
But that's precisely what CI is for :) Thanks.

PR #21, opened 2022-01-13
Dual red objects

blowfish880:
Rewrite the dual (hybrid) attack taking advantage of the new `short_vectors` feature of the reduction module.


malb:
Just as a sanity check, did you compare ADPS16 before and after and the results match?

blowfish880:
Good call. They don't. I'll try to figure out what's going on.

blowfish880:
Ah, I think I found it. It's a bug in [the previous version](https://github.com/malb/lattice-estimator/blob/main/estimator/lwe_dual.py#L143), it's taking `rho=4/3` instead of the square root. When changing that the results match.

malb:
Woohoo

PR #23, opened 2022-01-17
Kyber fixes

### LINKED ISSUE #22: Kyber reduction model too optimistic with many short vectors?
I know I ok'ed the code when it was merged, but looking at it again I think the Kyber model is too optimistic when estimating the cost of producing many short vectors. Specifically, the way it currently works is that `Kyber.short_vectors` calls `ReductionCost._short_vectors_sieve` with `beta-d4f(beta)`, which in turn calls `Kyber.__call__` with that beta. But `Kyber.__call__` will apply `d4f` again before estimating the complexity. So `d4f` is called twice leading to a much smaller `beta` and thus an overly optimistic estimate.

malb:
- additional factor `C` to cost in Kyber's cost model as per Kyber spec
- avoid double `d4f` in `Kyber.short_vectors`

Fixes #22

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,19 +21,19 @@ Quick Start
     LWEParameters(n=512, q=3329, Xs=D(σ=1.22), Xe=D(σ=1.22), m=512, tag='Kyber 512')
 
     >>> LWE.primal_usvp(Kyber512)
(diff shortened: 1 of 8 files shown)


blowfish880 (approved):
Looks good.

PR #28, opened 2022-01-25
Implement [GuoJoh21]

blowfish880:
This adds the attack from the recent Asiacrypt paper. As far as I can tell, the numbers we get are similar, but slightly worse (checked on Kyber only), even though the optimized parameters are very similar. Without public code it is hard to figure out where the difference comes from, so at this point I think this is as close as I'm gonna get without further input from the authors.

The paper adds two improvements:
1) A 2-step lattice reduction trick. This is already approximated in the Kyber model. This PR adds a new reduction class derived from Kyber, where the sieving step is performed with a slightly larger block size to balance the reduction and the sieving part. In some cases, this is slightly better.

2) This PR also adds an option to apply the FFT trick that allows to guess some coordinates mod 2. This is much slower, since it adds an additional parameter to optimize over, but can shave a couple of bits off from the estimate. It is off by default due to the increased running time.


blowfish880:
Looks like a test fails due to an output format issue. The tests run fine for me locally. Suggestions how to deal with that elegantly?

malb:
You can use ellipses: https://github.com/malb/lattice-estimator/blob/main/estimator/reduction.py#L274

estimator/lwe_dual.py
```diff
@@ -172,6 +185,56 @@ def cost(
         # just guess different components of the secret
         return cost.repeat(times=rep, select={"m": False})
 
+    @staticmethod
+    def gj21_solver(params, success_probability, t=0):
```
malb, line 189: should we call this "FFT" solver or so? Say, someone improves it, ideally we'd update this function rather than adding a new one (?)

estimator/lwe_dual.py
```diff
@@ -183,6 +246,7 @@ def optimize_blocksize(
         use_lll=True,
         log_level=5,
         opt_step=2,
+        gj21=False,
```
malb, line 249: Same as above

malb:
Looks great.!

Issue #20, opened 2022-01-13
Primal Hybrid gone wrong?


malb:
``` python
sage: D = ND.SparseTernary(1024, 256 // 2)
sage: LAC192v3a = LWE.Parameters(n=1024, q=251, Xs=D, Xe=D, m=1024, tag="LAC-192-v3a")
sage: LWE.primal_hybrid(LAC192v3a)
rop: ≈2^360.9, red: ≈2^359.9, svp: ≈2^359.9, β: 866, η: 2, ζ: 576, |S|: ≈2^510.8, d: 1003, prob: ≈2^-82.3, ↻: ≈2^84.5, tag: hybrid

sage: LWE.primal_hybrid(LAC192v3a, red_shape_model="CN11")
rop: ≈2^405.0, red: ≈2^405.0, svp: ≈2^378.1, β: 40, η: 2, ζ: 0, |S|: 1, d: 665, prob: ≈2^-357.2, ↻: ≈2^359.4, tag: hybrid
```

This doesn't look right.

malb:
It seems to boil down to `mitm_babai_probability` which perhaps doesn't like the tail shape (?) In that case, we should preprocess with Dual-BKZ to get a nice line in the critical lifting part?

blowfish880:
I personally believe that for decoding one should always reduce the dual basis. Then the GSA should be quite accurate in the tails and the result of using the simulator and the GSA should be fairly close, I would expect.

bencrts:
I've started to investigate this, will report back and open a PR asap.

bencrts:
I've made a first attempt to go via the dual basis on pull request #29. As the PR stands, this is the output for the same parameters as above:

```python
sage: from estimator import *
sage: sage: from estimator import *
sage: D = ND.SparseTernary(1024, 256 // 2)
sage: sage: LAC192v3a = LWE.Parameters(n=1024, q=251, Xs=D, Xe=D, m=1024, tag="LAC-192-v3a")

sage: sage: LWE.primal_hybrid(LAC192v3a)
rop: ≈2^361.7, red: ≈2^360.7, svp: ≈2^360.7, β: 863, η: 2, ζ: 576, |S|: ≈2^513.9, d: 1001, prob: ≈2^-81.6, ↻: ≈2^83.8, tag: hybrid

sage: sage: LWE.primal_hybrid(LAC192v3a, red_shape_model="CN11")
[...]

PR #29, opened 2022-02-09
bugfix(primal_hybrid): reduce via dual lattice for BKZ simulator

bencrts:
As in issue #20, we reduce via the dual lattice for the primal_hybrid attack. Changes include:

1. in `simulator.py` we add a `dual` flag to the `CN11` model which allows to go via the dual basis when performing simulation.
2.  in `lwe_primal.py` we add an `if/else` statement to allow using 1. 

Happy to discuss any details/API changes we want to make here.

### DIFF
--- a/estimator/lwe_primal.py
+++ b/estimator/lwe_primal.py
@@ -299,7 +299,12 @@ class PrimalHybrid:
 
         # 1. Simulate BKZ-β
         # TODO: pick τ
-        r = simulator(d, params.n - zeta, params.q, beta, xi=xi)
+
+        try:
+            r = simulator(d, params.n - zeta, params.q, beta, xi=xi, dual=True)
+        except TypeError:
+            r = simulator(d, params.n - zeta, params.q, beta, xi=xi)
+
         bkz_cost = costf(red_cost_model, beta, d)
 
         # 2. Required SVP dimension η
--- a/estimator/simulator.py
+++ b/estimator/simulator.py
@@ -20,7 +20,7 @@ The last row is optional.
 from sage.all import RR, log
 
 
-def CN11(d, n, q, beta, xi=1, tau=1):
+def CN11(d, n, q, beta, xi=1, tau=1, dual=False):
     from fpylll import BKZ
     from fpylll.tools.bkz_simulator import simulate
 
@@ -29,7 +29,20 @@ def CN11(d, n, q, beta, xi=1, tau=1):
     else:
         r = [q ** 2] * (d - n) + [xi ** 2] * n
 
-    return simulate(r, BKZ.EasyParam(beta))[0]
+    if dual is False:
+        return simulate(r, BKZ.EasyParam(beta))[0]
+
+    else:
+        # 1. reverse and reflect the basis (go to dual)
+        r = [1/r_ for r_ in r[::-1]]
+
+        # 2. simulate reduction on the dual basis
+        r = simulate(r, BKZ.EasyParam(beta))[0]
+
+        # 3. reflect and reverse the basis (go back to primal)
+        r = [1/r_ for r_ in r[::-1]]
+
+        return r
 
 
 def GSA(d, n, q, beta, xi=1, tau=1):


estimator/simulator.py
```diff
@@ -29,7 +29,25 @@ def CN11(d, n, q, beta, xi=1, tau=1):
...
+
+        # 2. simulate reduction on the dual basis
+        r = simulate(r, BKZ.EasyParam(beta))[0]
+
+        # 3. reflect and reverse the basis (go back to primal)
+        r_log = [log(s, 2) for s in r]
+        r_ref = [-s for s in r_log]
+        r_log.reverse()
+        r = [2**s for s in r_ref]
+        return r
```
bencrts, line 51: Not sure if this conversion to/from logs is the most efficient.

estimator/lwe_primal.py
```diff
@@ -299,7 +299,10 @@ def cost(
 
         # 1. Simulate BKZ-β
         # TODO: pick τ
-        r = simulator(d, params.n - zeta, params.q, beta, xi=xi)
+        if simulator is CN11:
+            r = simulator(d, params.n - zeta, params.q, beta, xi=xi, dual=True)
+        else:
+            r = simulator(d, params.n - zeta, params.q, beta, xi=xi)
         bkz_cost = costf(red_cost_model, beta, d)
```
bencrts, line 306: We might be able to remove this `if/else` statement, but I couldn't figure out the best way to do it.

malb: Can't you just do
```
try:
  f(dual=True)
except TypeError:
 f()
```
?

estimator/simulator.py
```diff
@@ -29,7 +29,25 @@ def CN11(d, n, q, beta, xi=1, tau=1):
...
-    return simulate(r, BKZ.EasyParam(beta))[0]
+    if dual is False:
+        return simulate(r, BKZ.EasyParam(beta))[0]
+
+    else:
+        # 1. reverse and reflect the basis (go to dual)
+        r_log = [log(s, 2) for s in r]
+        r_log.reverse()
+        r_ref = [-s for s in r_log]
+        r = [2**s for s in r_ref]
```
malb, line 40: Why not `r = [1/r_ for r_ in r[::-1]]`?

estimator/lwe_primal.py
blowfish880: How about just adding the `dual` flag to the simulator interface? I always cringe when a function needs to know which specific instantiation of some type it is given. GSA can just ignore the flag then (but maybe add a comment why). 

Another way would be to simply do the reverse and invert of the basis in both cases as part of a generic "dual_simulate" function that may be applied to any simulator. That would uselessly apply the transformations also in the GSA case, so there might be some slow down, but if it's not too bad, that might be a nice solution, since it makes it easy to have new simulators that are compatible with the primal hybrid.

malb: I like the `dual_simulate` approach. Will need to extract some LLL `r` shape generically for `d,k,q,xi` but that's perhaps useful more generally, too.

malb: Ah, the annoying thing with `dual_simulate(simulator, d, params.n - zeta, params.q, beta, xi=xi)` is that the simulators don't expect `r` but `d,k,q,xi`. So perhaps the right thing™ is to rewrite the simulator interface to take `r` as input and then we provide our own higher level functions that take `d,k,q,xi` and produce GSO norms. These can then also transparently handle the dual case, simulators don't need to know. To avoid the slow down, we could special case the GSA thing inside these higher level functions?

blowfish880: Yes, I agree that makes sense. This sounds like a bit more work, so if @bencrts  doesn't wanna do this refactoring right now, we could approve this PR as is and leave this for a separate PR.

malb: Agreed

estimator/simulator.py
bencrts: Fixed!

estimator/lwe_primal.py
bencrts: I agree that the suggested changes (e.g. `dual_simulate()`) are probably a better idea. Happy to start working towards this at some point next week.

I've fixed all of the outstanding issues in the code review, and I agree that merging is a good idea (that way we get the fix into the codebase asap).

PR #31, opened 2022-02-10
dual simulator

### LINKED ISSUE #30: Generically support reduction of the dual basis

malb:
fixes #30

### DIFF
--- a/estimator/simulator.py
+++ b/estimator/simulator.py
@@ -83,8 +83,8 @@ def GSA(d, n, q, beta, xi=1, tau=1, dual=False):
     :param beta: Block size β.
     :param xi: Scaling factor ξ for identity part.
     :param tau: Kannan factor τ.
-    :param dual: ignored
-
+    :param dual: ignored, since GSA is self-dual: applying the GSA to the dual is equivalent to
+           applying it to the primal.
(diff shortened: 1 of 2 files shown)


estimator/simulator.py
```diff
@@ -20,41 +20,70 @@
...
+    """
+    Reduced lattice shape following the Geometric Series Assumption [Schnorr03]_
 
     :param d: Lattice dimension.
     :param n: Number of `q` vectors
     :param q: Modulus `q`
     :param beta: Block size β.
     :param xi: Scaling factor ξ for identity part.
     :param tau: Kannan factor τ.
+    :param dual: ignored
```
blowfish880, line 86: Maybe add a comment along the lines of "ignored since applying the GSA to the dual is equivalent to applying it to the primal" so it doesn't look like it's not supported?

blowfish880 (commented):
Looks good. So this allows now to remove the [try-except block](https://github.com/malb/lattice-estimator/blob/main/estimator/lwe_primal.py#L303) from `lwe_primal.py` I guess. Should this be part of this PR as well?

malb:
Cool, changed accordingly.

PR #40, opened 2022-08-25
Fix calculation of stddev for SparseTernary. fixes #39

### LINKED ISSUE #39: Question about stddev of SparseTernary
Hi,
I'm a bit confused by how you calculate the standard deviation for your SparseTernary distribution.
You calculate it as `sqrt( (p + m) / n )`, but doesn't that only apply when the mean is zero? i.e. when `p == m`?
I'd like to use your tool with a binary key and I calculated the standard deviation for a binary key as `sqrt( ( hn - h^2 ) / n^2 )`.
The formula for the general case when `p != m` should also be different.
Am I misunderstanding something or is the code assuming that `p == m`?

3553x:


malb:
Thanks!

PR #41, opened 2022-09-08
Document LWEParameters member variables

j2kun:
Hi there 👋 

I'm researching a bit about LWE security, and as I'm digging through the codebase and paper, I figured I might as well add some extra documentation and similar quality-of-life improvements to the repository. I _believe_ I understood the semantics of the parameters in this class, but if I'm wrong please let me know and I will fix it!

Also let me know if you have any specific requests for doc style that I should follow. For example, should I put TeX source in the sphinx autodocs? Or perhaps in code fences? (see `Z/qZ` and similar below, which is currently plain text)

This PR also adds two standard lines to the `.gitignore`.

Here is what the new sphinx documentation looks like after this change

image

### DIFF
--- a/estimator/lwe_parameters.py
+++ b/estimator/lwe_parameters.py
@@ -8,12 +8,18 @@ from .errors import InsufficientSamplesError
 
 @dataclass
 class LWEParameters:
-    n: int
-    q: int
-    Xs: NoiseDistribution
-    Xe: NoiseDistribution
+    """The parameters for a Learning With Errors problem instance."""
+
+    n: int  #: the dimension of the LWE sample vector (Z/qZ)^n.
+    q: int  #: the modulus of the space Z/qZ of integers the LWE samples are in.
+    Xs: NoiseDistribution  #: the distribution on Z/qZ from which the LWE secret is drawn
+    Xe: NoiseDistribution  #: the distribution on Z/qZ from which the error term is drawn
+
+    #: the number of LWE samples allowed to an attacker,
+    #: optionally `sage.all.oo` for allowing infinitely many samples.
     m: int = oo
-    tag: str = None
+
+    tag: str = None  #: a name for the patameter set
 
     def __post_init__(self, **kwds):
         self.Xs = copy(self.Xs)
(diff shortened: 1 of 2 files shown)


malb:
Thanks! Good question for maths typesetting. At the moment I put no guards around it and just use unicode where available. Perhaps the right choice is to do "`Z/qZ`" which, IIRC, Sphinx can be convinced to interpret as math environments?

estimator/lwe_parameters.py
```diff
@@ -8,12 +8,16 @@
...
+
+    n: int  #: the dimension of the LWE sample vector (Z/qZ)^n.
+    q: int  #: the modulus of the space Z/qZ of integers the LWE samples are in.
+    Xs: NoiseDistribution  #: the distribution on Z/qZ from which the LWE secret is drawn
+    Xe: NoiseDistribution  #: the distribution on Z/qZ from which the error term is drawn
+
+    #: the number of LWE samples allowed to an attacker,
+    #: optionally `sage.all.oo` for allowing infinitely many samples.
     m: int = oo
-    tag: str = None
```
malb, line 16: We needs tag

j2kun:
Fixed the missing `tag` (it was there but got removed in a bad rebase).

I imagine the tests will pass now, but I'm unable to run the tests myself locally. I have sage installed, and a virtualenvironment with `pip install requirements.txt`, but it can't seem to be able to find the `sage` module when I run `pytest`. I imagine this is because it's not running from the `sage` interpreter directly, but then running something like `sage -python` is not using the virtual environment at all.

Are you installing from pip like here? https://wiki.sagemath.org/ReleaseTours/sage-9.3#Alternative_installation_methods_using_pip

malb:
Is pytest installed in the same virtualenv as SageMath? FWIW I run Sage inside `conda` which works quite well for me.

j2kun:
That's the core of my confusion, because sagemath can't be installed via pip, though I have not tried the method from my previous comment yet

https://wiki.sagemath.org/ReleaseTours/sage-9.3#Alternative_installation_methods_using_pip

I will give it a few more shots.

malb:
Note that there is `sage -pip` and `sage -sh` which might get you where you want to be

malb:
Thanks! You should also open a PR to add yourself to the list of contributors.

PR #43, opened 2022-09-11
Add call to register_impermanent in gb.py. fixes #42

### LINKED ISSUE #42: Bug "function does not know about 'dreg'"
Running the following code results in a bug:
```python
Xs = ND.SparseTernary(1024, 64, 0)
Xe = ND.DiscreteGaussian(2**10)
params = LWE.Parameters(n=1024, q=2**40, Xs=Xs, Xe=Xe)
LWE.estimate(params)
```

This is because a call to `register_impermanent` is missing in `gb.py`.

3553x:
This commit fixes #42

### DIFF
--- a/estimator/gb.py
+++ b/estimator/gb.py
@@ -224,6 +224,7 @@ class AroraGB:
         params = params.normalize()
 
         best = Cost(rop=oo, dreg=oo)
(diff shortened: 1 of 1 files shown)


malb:
Thanks!

PR #44, opened 2022-09-22
Add Dockerfiles

j2kun:
This will help ensure I can reliably run the project's tests.

However, I did notice that the tests are very slow (with or without docker). What is the expected runtime of the tests? In particular, I see the following when running the `README.rst` tests

```bash
time docker exec lattice-estimator-tests sage -sh -c pytest -k README
<...>
0.33s user 0.40s system 0% cpu 4:50.98 total
```

(the whole test suite takes 21 minutes on my setup)

I suspect it could be sage startup time being incurred on each test case?

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -1,0 +1,9 @@
+# A Dockerfile that can be used in isolation
+# syntax=docker/dockerfile:1
+FROM sagemath/sagemath:9.5
+
+COPY --chmod=777 . /lattice-estimator
+WORKDIR "/lattice-estimator"
+
+RUN sage -pip install -r requirements.txt
+CMD ["sage"]
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -1,5 +1,8 @@
-# A Dockerfile that can be used in isolation
 # syntax=docker/dockerfile:1
+
+# A Dockerfile that uses the local clone of lattice-estimator, incorporating
+# local changes that have not yet been merged into malb/lattice-estimator at
+# HEAD.
 FROM sagemath/sagemath:9.5
 
 COPY --chmod=777 . /lattice-estimator
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,19 @@
+# Running lattice estimator in a container
+
+Requires `docker` or `podman` to run the containers.
+
+From the base of the repository, run the following to build an image and run the
+tests within the container.
+
+```bash
+docker build -t lattice-estimator -f docker/Dockerfile.dev .
+docker run -dit --name lattice-estimator-tests lattice-estimator
+docker exec lattice-estimator-tests sage -sh -c pytest
+```
+
+Note that due to [this open
+ticket](https://trac.sagemath.org/ticket/34242#comment:20) on Sage, the
+published sage container is using an OEL'ed version of Ubunutu, and so we have
+to wait for the sagemath image to be updated in order to use `Dockerfile`
+standlone (`git clone`ing from inside the container).
+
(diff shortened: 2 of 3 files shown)


j2kun:
Ah, silly question, I see from the last PR that the runtime should be ~8 minutes

https://github.com/malb/lattice-estimator/pull/41/checks

I'll have to investigate why my run is so much slower.

malb:
I love the idea! But those requirements seem overkill, can we split them out?

j2kun:
I'm not entirely sure what you mean about the requirements being overkill, and at the risk of explaining something you already know, let me say what I did for 81be90f21d985aaaa3180cfe077a5b44111bfd3c. tl;dr: No new requirements were added, I just used `pip freeze` after installing the existing requirements.

The previous list of requirements had two features that one might consider "problems" (if one is, like me, passionate about stability and visibility in software):

1. The versions of the requirements were not pinned.
2. The transitive dependencies of the listed requirements were not shown.

The second problem is minor, but listing all requirements can be helpful to see if any large or unnecessary packages are being pulled in transitively, which could point out opportunities to simplify and speed up a build/test pipeline.

The first one is more of a problem because a dependent package can push a new version which (unbeknownst to the package publisher) breaks this project. You might not notice it until (a) a new user comes along and complains the project is broken or (b) a PR fails tests as a result---either because the install process fails, or because the subtle change in behavior of a dependency causes a test to mysteriously fail in a way that is unrelated to the change in the PR (a debugging nightmare!). 

Failures due to silent upstream updates is usually a distraction, and constantly using the latest version of a given package is usually not necessary and need only be done to incorporate a critical bug fix from upstream. Pinning the versions of all the requirements ensures that you get consistent behavior, and you can still manually upgrade any given package and test it in isolation. (`pip install -U package`, then `pip freeze > requirements.txt`, then create a new PR with just that change).

So, the process I used to achieve this is simply:

1. Run `pip install -r requirements.txt` (rather, using `sage -pip`)
2. Run `pip freeze > requirements.txt`

j2kun:
Ah, I think you are asking me to put the change to the requirements in a separate PR?

malb:
Mhhh, I get your reasoning for production code, but this is a research library where I don't want audit dependency version numbers for vulnerabilities/bugs. But maybe we can move this to a separate PR and discuss there?

j2kun:
Split requirements change into https://github.com/malb/lattice-estimator/pull/46

docker/README.md
```diff
@@ -0,0 +1,19 @@
...
+docker build -t lattice-estimator -f docker/Dockerfile.dev .
+docker run -dit --name lattice-estimator-tests lattice-estimator
+docker exec lattice-estimator-tests sage -sh -c pytest
+```
+
+Note that due to [this open
+ticket](https://trac.sagemath.org/ticket/34242#comment:20) on Sage, the
+published sage container is using an OEL'ed version of Ubunutu, and so we have
+to wait for the sagemath image to be updated in order to use `Dockerfile`
+standlone (`git clone`ing from inside the container).
```
malb, line 18: standalone?

docker/Dockerfile.dev
```diff
@@ -0,0 +1,9 @@
+# A Dockerfile that can be used in isolation
```
malb, line 1: What is Dockerfile.dev for?

j2kun: This comment is a mistake from copy/paste. I added a second commit explaining the use cases of the two Dockerfiles. In short, `Dockerfile` always uses malb/lattice-estimator at HEAD, and `Dockerfile.dev` uses a local copy of the repository, including any un-committed changes.

docker/README.md
j2kun: Once that issue is fixed, one can just copy `Dockerfile` to their local system, run it, and use `lattice-estimator` from within the container. They don't have to clone the repository or set up sage on their system.

PR #47, opened 2022-09-23
Update test running command in contributor docs

cathieyun:
The existing test command `python -m pytest` doesn't correctly run tests because sage can't be imported into python; instead, the tests should be run from the sage environment.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -68,7 +68,7 @@ We explain what is going on above:
 Tests
 -----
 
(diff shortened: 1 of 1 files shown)


malb:
Good catch, I'd say it should be `pytest`, `python -m pytest` or `sage --python -m pytest`? The second one for those running inside `conda` or `sage -sh`?

malb:
Thanks! Add yourself to the list of contributors too!

PR #49, opened 2022-10-05
Add TFHE parameters from the Concrete library

cathieyun:
The Concrete library uses parameters that were chosen for faster boolean circuit evaluation. Add these parameters to the set of schemes for the lattice estimator to evaluate.

Concrete parameter set:
https://github.com/zama-ai/concrete/blob/main/concrete-boolean/src/parameters/mod.rs#L83

### DIFF
--- a/estimator/schemes.py
+++ b/estimator/schemes.py
@@ -201,6 +201,19 @@ TFHE1024 = LWEParameters(
     tag="TFHE1024",
 )
 
+# CONCRETE default parameter set for 128-bit security, chosen for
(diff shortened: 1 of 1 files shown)


malb:
Shouldn't we tag these as "Concrete-TFHE586" or so?

cathieyun:
Sure, I didn't know if you had a preferred naming scheme. I'll update the name.

bencrts (approved):
Looks good to me! We’re in the process of updating these parameter sets, but I can create a new PR when we make the change public in Concrete.

Issue #48, opened 2022-10-04
Where does RLWE fit into this repository?


j2kun:
I'm curious how the RLWE scheme fits into the attacks in this project. I've seen a number of papers where authors choose parameters for RLWE/RGSW, and the error distributions are usually much smaller than the corresponding LWE parameters. I'm curious how those are justified, and whether this project is expecting to add support for RLWE-based estimates.

malb:
It's actually the other way around. We typically cost attacks on RLWE by pretending we just have LWE instances in front us. That is, we do not know of algorithms that exploit the algebraic structure in RLWE on top of LWE (some minor caveats apply, but the statement holds for parameters typically considered). I assume those parameter difference are simply explained by those RLWE instances having larger dimensions which permits to use smaller noise, perhaps?

j2kun:
Usually the RLWE dimensions is 1, but the degree of the polynomials are larger (e.g., for TFHE, lwe_dim=630, rlwe_dim=1, rlwe_polymod_deg=1024). In that case, an RLWE instance (two polys in a sample each with degree 1024) corresponds to an LWE instance with dimension... 2048? 1024? 2?

What is that mapping from RLWE parameters to LWE parameters? I skimmed Rachel Player's thesis looking for hints but didn't find any... if you have a reference, I'd be happy to read it and add that mapping as a function in this repo for future bozos like me :)

malb:
Ah, yeah, so if your ring is e.g. `R = ZZ[x]/<x^d + 1>` for d a power of two, then the corresponding LWE dimension is d. For a in R you get a d x d matrix A where the rows (or columns depending of how you look at it) are (a, x*a % (x^d + 1), x^2*x % (x^d +1), ...)

j2kun:
Summarizing from https://github.com/malb/lattice-estimator/pull/51

For a "vanilla" RLWE the degree of x^N + 1 is the corresponding LWE dimension. Implementations of the reductions for specific RLWE variants, like CGGI's TRLWE, are not in scope of this repository, and the "vanilla" reduction is too trivial and restrictive to warrant explicit code.

PR #51, opened 2022-10-12
Add RLWEParameters and reduction to LWE

j2kun:
Fixes https://github.com/malb/lattice-estimator/issues/48

Hi Martin,

I added the simple reduction from RLWE to LWE in an RLWEParameters class. I wanted to get early feedback from you on aligning with your contributing guidelines before, e.g., adding constants for specific RLWE parameter instances in the literature (or adding tests, looks like it's all doctests?).

In particular, I ran into a roadblock when generating the docs (to see how my formatted matrices look and tweak them), and I ran into this error:

```
WARNING: [autosummary] failed to import estimator.rlwe_parameters.
Possible hints:
* ImportError: no module named estimator
* TypeError: '<' not supported between instances of 'int' and 'MagicMock'
* KeyError: 'estimator'
```

I think this is happening because of [the MagicMock in docs/conf.py](https://github.com/malb/lattice-estimator/blob/main/docs/conf.py#L37), which is being compared to an int in `__post_init__`; however, I don't see how this is not causing the same problem in `lwe_parameters.py`. I'm not particularly familiar with sphinx (why would it be running this code to run into the type error instead of statically analyzing it?), but I also noticed that `sphinx-book-theme` is preventing sphinx version >= 5, which I imagine might have bug fixes that would avoid this: https://github.com/executablebooks/sphinx-book-theme/issues/592.

Open to any suggestions you have to fix this!

### DIFF
--- a/docs/api_doc.rst
+++ b/docs/api_doc.rst
@@ -21,6 +21,7 @@ API Reference
    estimator.nd
    estimator.prob
    estimator.reduction     
+   estimator.rlwe_parameters
    estimator.simulator
    estimator.util
 
--- a/estimator/rlwe_parameters.py
+++ b/estimator/rlwe_parameters.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+import math
+from sage.all import oo
+from dataclasses import dataclass
+from copy import copy
+
+from .lwe_parameters import LWEParameters
+from .nd import NoiseDistribution
+
+
+@dataclass
+class RLWEParameters:
+    """The parameters for a Ring Learning With Errors problem instance."""
+
+    N: int  #: The degree of the polynomial modulus x^N + 1
+    q: int  #: The modulus of the space Z/qZ the coefficients are in.
+    Xs: NoiseDistribution  #: the distribution on Z/qZ from which the LWE secret is drawn.
+    Xe: NoiseDistribution  #: the distribution on Z/qZ from which the error term is drawn.
+
+    #: the number of LWE samples allowed to an attacker,
+    #: optionally `sage.all.oo` for allowing infinitely many samples.
+    m: int = oo
+
+    tag: str = None  #: a name for the patameter set
+
+    def __post_init__(self, **kwds):
+        self.Xs = copy(self.Xs)
+        self.Xs.n = 1  # RLWE instances are a dot product of two polynomials.
+
+        log_N = math.log2(self.N)
+        if not log_N.is_integer():
+            raise ValueError(
+                "RLWE security estimation only supports power-of-2 polynomial "
+                f"modulus degrees, but {self.N} (log2={log_N}) was given to "
+                "RLWEParameters.")
+
+        if self.m < oo:
+            self.Xe = copy(self.Xe)
+            self.Xe.n = self.m
+
+    def as_lwe_params(self) -> LWEParameters:
+        """Reduce this instance of RLWE to LWE.
+
+    This operates under the assumption that no current known attacks exploit the
+    ring structure of RLWE beyond its LWE structure.
+
+    The reduction starts with an RLWE instance with sample
+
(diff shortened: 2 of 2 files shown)


j2kun:
Oh, one other question I had about this reduction: wouldn't a reduction from RLWE to LWE require that the resulting LWE samples are independent? In this case, the samples that arise from the RLWE instance are correlated, no?

j2kun:
On further thought, for RLWE with vectors of polynomials, wouldn't you just stack the dimension-1 reduction horizontally, and then repeat the secret key vector `dim` many times? Somehow that seems like is will violate something about the distributions, because the secret key is no longer uniform...

j2kun:
Also, on the doc generation questions, it looks like the autodoc generation is [failing in the github action](https://github.com/malb/lattice-estimator/actions/runs/3238332644/jobs/5306440459#step:4:205) as well, though it has an import error for all the modules, not just my new one...

malb:
Hi Jeremy, let's talk about the mathematics/cryptography separately from Python issues. I don't think the PR works as is. Ring-LWE is not only defined for power-of-two ring dimensions. It is defined for, at least, any cyclotomic ring, e.g. `sage: CyclotomicField(127).polynomial()`. On some level, there is then no translation: parameters just translate one to one and I don't think a new class is needed. If we wanted to be precise, there would be differences (the spaces where elements live, cf. canonical embeddings) but those don't (seem to!) matter for security estimates. So I'm afraid adding something like this would only add confusion.

malb:
If we're given samples `(a_i, a_i * s + e_i)` we can consider `(A_i, A_i*s + e_i)` where `A_i = rot(a_i)` the rotational matrix of the ring element `a_i`. So each RLWE sample simply gives you `n` LWE samples. Yes, these are correlated, but as far as we know this doesn't seem to matter to algorithms.

malb:
I'm at a loss on the Sphinx question, though.

j2kun:
My intention with this PR was mainly to document the provenance of the LWE parameter sets coming from RLWE instantiations, since those are not clearly explained in the cited papers. For example, the [TFHE1024 set](https://github.com/malb/lattice-estimator/blob/main/estimator/schemes.py#L254) marks n=1024 in LWE parameters, but the paper it comes from uses RLWE and is at best vague about the relationship. This is all they have to say: 

image

And then in the table cited by lattice-estimator, the authors silently switch from `N=1024, k=1` (RLWE with only one sample polynomial) to `n=1024` and claim security. I recall other parameter sets that used `N=512, k=2` (512-degree with two sampled polynomials), where it seems the product of those two parameters gives you the corresponding LWE parameter set. For me, it would be clearer to see, e.g., `RLWEParameters(N=512, k=2).as_lwe_params()` and know that there is a conversion happening, albeit a very simple one, than to see `LweParameters(n=1024)` with no further explanation. In addition, if "RLWE" includes the "generalized" RLWE in the CGGI line of work, then the N=512, k=2 example shows it's not 1-1.

I have also heard that, if N is not a power of 2 (and/or the secret is not binary) then the RLWE distributions don't reduce to 
LWE neatly (or at all?), so in that case it seems lattice-estimator can't be applied. The "guards" via raised ValueErrors would serve as documentation of that folk lore. In addition to what you mentioned, that these reductions are applicable mainly because we don't have any stronger attacks that rely on the ring structure (and the correlation of samples).

Happy to accept the final call and close the PR if you're still unconvinced.

malb:
I suspect by `N=512, k=2` people mean M-LWE with rank `k=2` and ring dimension `N=512`, i.e. overall secret dimension `1024`? It is true that when `N` is not a power of two you may have to worry about the distortion of the noise when considering it in the "coefficient embedding". However, in the "canonical embedding" this issue doesn't exist.

I'm afraid I still think this PR would rather add confusion than make things more accessible.

PR #60, opened 2022-12-01
Perform a sweep across lattice parameters, and graph the results

cathieyun:
This performs a parameter sweep across a range for `n` and `sigma` at specified intervals (and using `log_2(sigma)` in practice). It then graphs the resulting security for LWE instances with those lattice parameters.

This can be a useful tool for cryptographers who want to choose lattice parameters that have a certain security, and want an easy way to visualize the parameter space. It seems that multiple research/development teams have used a similar approach [1][2], so adding this parameter sweep script to the lattice estimator can reduce duplicated work.

Example output images, from running the following command:
`sage --python param_sweep.py 600:1140:20 6:20:1`

Sweeping from n=[600, 1140) by increments of 20, and Xe=[6, 20) by increments of  1, gradient:
28-11-2022_22-47-18_n600-1140-20_xe6-20-1_gradient

Sweeping from n=[600, 1140) by increments of 20, and Xe=[6, 20) by increments of  1, cutoff at 128 bits of security:
28-11-2022_22-47-18_n600-1140-20_xe6-20-1_cutoff

[1] Hitchhiker's Guide to TFHE: https://eprint.iacr.org/2022/1315.pdf
[2] Estimating the Security of Homomorphic Encryption Schemes: https://medium.com/zama-ai/estimating-the-security-of-homomorphic-encryption-schemes-cb798f9378f


malb:
I'd suggest to generalise it a bit, e.g.:

```python
from sage.all import log
from estimator import *

def security_level(params, f=LWE.estimate, *args, **kwds):
    res = f(params, *args, **kwds)
    return min([log(res_.rop, 2) for res_ in res.values()])

def parameter_sweep(n, q, sigma_e, sigma_s, m=None, Xe=ND.DiscreteGaussian, Xs=ND.DiscreteGaussian, tag=None, f=LWE.estimate, *args, **kwds):
    try:
        iter(n)
    except TypeError:
        n = [n]
    try:
        iter(q)
    except TypeError:
        q = [q]
    try:
        iter(m)
    except TypeError:
        m = [m]
    try:
        iter(sigma_e)
    except TypeError:
        sigma_e = [sigma_e]
    try:
        iter(sigma_s)
    except TypeError:
        sigma_s = [sigma_s]

    results = {}
        
    for n_ in n:
        for q_ in q:
            for sigma_e_ in sigma_e:
                for sigma_s_ in sigma_s:
                    for m_ in m:
                        if m_ is None:
                            m_ = n_
                        params = LWE.Parameters(n=n_, q=q_, Xe=Xe(sigma_e_), Xs=Xs(sigma_s_), m=m_)
                        lvl = security_level(params, f=f, *args, **kwds)
                        results[(n_,q_,sigma_e_,sigma_s_,m_,tag)] = lvl
    return results

res = parameter_sweep(n=range(300,401,50), q=[3329, 7681], sigma_e=2.0, sigma_s=2.0, f=LWE.estimate.rough)
```

cathieyun:
I updated the code to generalize it for `n`, `q`, `Xe`, and `Xs` as you suggested. 

It only makes sense to have 2 or fewer variables at a time for the graph utility though. For two variables, we can do a 2D heat graph - but for more variables, there isn't a great way to have a 3D visualization of a heat graph (which is the best way to visually represent that output - besides having multiple 2D heat graphs, which isn't very easy to use). So I just capped the graphing utility to 2 variables, though if a user wants to iterate over more variables, they can still do so and print the output dictionary from `result_dict`.

param_sweep.py
```diff
@@ -0,0 +1,374 @@
...
+        # value is a list
+        arg_val = [float(val) if '.' in val else int(val) 
+            for val in arg_val.split(',')]
+        return arg_val
+    else:
+        # value is a single value
+        return [float(arg_val) if '.' in arg_val else int(arg_val)]
+
+
+def GraphResults(result_dict: dict,
```
malb, line 100: This should be in `lower_case` to be consistent with the rest of the code.

param_sweep.py
```diff
@@ -0,0 +1,374 @@
...
+
+            ax2.set_xlabel(f'Parameter: {x_param}')
+            ax2.set_ylabel(f'Parameter: {y_param}')
+
+            fig2.savefig(file_name + '_cutoff.png')
+            Logging.log('sweep', log_level, 'Saved the cutoff graph to: %s',
+                        file_name + '_cutoff.png')
+
+
+def PerformCalculation(input_params: tuple[int, float],
```
malb, line 218: This should be `perform_calculation`

malb: Maybe `estimatef`  or so? To make it clearer what this function does?

param_sweep.py
```diff
@@ -0,0 +1,374 @@
...
+    print("params: ", params)
+    results[input_params] = security_level(params, f=f_)
+
+
+def security_level(params, f=LWE.estimate, *args, **kwargs):
+    res = f(params, *args, **kwargs)
+    return min([math.log(res_.rop, 2) for res_ in res.values()])
+
+
+if __name__ == "__main__":
```
malb, line 247: Is a command line script a good idea here? Wouldn't it be more flexible to call this from iPython? This would also simplify the code?

param_sweep.py
```diff
@@ -0,0 +1,374 @@
...
+    Creates one or more graphs, depending on how many variables are being
+    iterated over.
+
+    In the case of a single variable being a range, the output is a simple plot.
+    In the case of two variables, a heatmap plot is created.
+    In the case of three variables, a collection of heatmap plots are created.
+    For four or more variables, an error is raised, as it is not practical to
+    convey that information in a usable plot.
+
+    Args:
```
malb, line 115: Can you change the params format to match the other code?

param_sweep.py
```diff
@@ -0,0 +1,374 @@
...
+    if not output_dir:
+        output_dir = os.path.dirname(os.path.realpath(__file__))
+    assert args.num_proc >= 1, 'need at least one process to execute'
+    file_name = time.strftime('%d-%m-%Y_%H-%M-%S')
+    file_name = os.path.join(output_dir, file_name)
+
+    if not args.load_pickle:
+        result_dict = multiprocessing.Manager().dict()
+        work = []
+        for n_ in n:
```
malb, line 337: I think this should be a standalone function: take the Cartesian product of these parameter ranges and produce a dictionary of the results, possibly in parallel. This could be useful in other contexts too.

param_sweep.py
```diff
@@ -0,0 +1,374 @@
...
+
+            for (j, i), label in np.ndenumerate(mat):
+                ax.text(i, j, round(label, 1), ha='center', va='center', color='white')
+            plt.title(f'Security with parameters {desc_params}'
+                      f' and cutoff at {security_cutoff}')
+
+            ax2.set_xlabel(f'Parameter: {x_param}')
+            ax2.set_ylabel(f'Parameter: {y_param}')
+
+            fig2.savefig(file_name + '_cutoff.png')
```
malb, line 213: should we allow the user to provide a filename including extension? People might want a PDF or SVG?

param_sweep.py
cathieyun: Sounds good. I put the functionality of doing a parameter sweep (without graphing) into a separate function, `parameter_sweep`.

param_sweep.py
cathieyun: That does make much more sense, I updated it the code to be callable from iPython / sage!

param_sweep.py
cathieyun: Done - I combined this with the `security_level` function and just kept the `security_level` name.

param_sweep.py
cathieyun: Added that functionality!

cathieyun:
I did a preliminary pass to address the feedback. I'll update the comments / documentation also, soon!

cathieyun:
I cleaned up the code a bit - I added documentation, examples, and type annotations. If this looks good, feel free to merge the PR!

malb:
Hi @cathieyun, looks good! 

I only have bikeshedding comments:
- we use a different convention for writing parameters in docstrings, see e.g. https://github.com/malb/lattice-estimator/blob/main/estimator/lwe.py#L37
- we use a different convention for giving examples, which we also doctest by default, see e.g. https://github.com/malb/lattice-estimator/blob/main/estimator/lwe.py#L41
- I'd personally not have `ParameterSweep` and especially `TestParameterSweep` and just have the functions at the top level (I guess `security_level` would go into `lwe.py`?)

cathieyun:
Addressing the comments:
* I changed the formatting of the parameters in the docstrings, to match the existing code
* I moved the tests to the doctest format. This makes them more brittle (if the lattice estimator changes its security estimate for a given parameter set, the tests now need to be changed) but this seems to be true for the rest of the doctests as well.
* By moving the tests to doctests, I got rid of `TestParameterSweep`. As for `ParameterSweep`, I do think the functions should live under this class, in order to keep the code well organized / compartmentalized, so I kept that class.

malb:
Looks good to me. Is it ready to merge from your end?

malb:
Ta!

PR #61, opened 2022-12-01
Don't export any scheme names.

joerowell:
At the moment there's a mismatch in how to use different schemes:

```
sage: from estimator import *
sage: Kyber512
LWEParameters(n=512, q=3329, Xs=D(σ=1.22), Xe=D(σ=1.22), m=512, tag='Kyber 512')
sage: Frodo640
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-3-c02c65839524> in <module>
----> 1 Frodo640

NameError: name 'Frodo640' is not defined
```

Instead, one has to specify ```schemes``` first:

```
sage: schemes.Frodo640
LWEParameters(n=640, q=32768, Xs=D(σ=2.80), Xe=D(σ=2.80), m=656, tag='Frodo640')
```

This PR simply removes all of the schemes in ```__init__.py```  to make everything uniform.


malb:
I agree the asymmetry is not nice. I wonder if we should in the opposite direction and add none to the global namespace? Any preferences?

joerowell:
Adding none is probably nicer, as then we don't need to remember to add them to ```__init__.py```.  On the other hand, it probably breaks any scripts people have already written against the estimator for Kyber etc...

malb:
Not in the business of providing backward compatibility :)

malb:
Thank you!

PR #62, opened 2022-12-02
Add schemes back to __init__.py.

joerowell:
Bizarrely, the tests pass but (at present) this code fails:

```
sage: from estimator import *
sage: schemes.Frodo640
... 
NameError: name 'schemes' is not defined.
```

This PR fixes this by adding an import for schemes to __init__.py. No idea why this wasn't caught by the CI.


malb:
Ta!

Issue #63, opened 2022-12-07
dual_hybrid crash


bencrts:
`dual_hybrid` is crashing when we choose `red_cost_model = RC.Kyber`, e.g.

```
sage: from estimator import *
sage: LWE.dual_hybrid(schemes.TFHE630, red_cost_model=RC.Kyber)
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/mambaforge/envs/sage/lib/python3.8/site-packages/sage/misc/cachefunc.pyx in sage.misc.cachefunc.CachedFunction.__call__ (build/cythonized/sage/misc/cachefunc.c:5959)()

[...]

~/Desktop/Github/lattice-estimator/estimator/lwe_dual.py in cost(solver, params, beta, zeta, h1, t, success_probability, red_cost_model, use_lll, log_level)
    144
    145         # only care about the scaling factor and don't know d yet -> use 2 * beta as dummy d
--> 146         rho, _, _, _ = red_cost_model.short_vectors(beta=beta, d=2 * beta)
    147
    148         params_slv, m_ = DualHybrid.dual_reduce(

ValueError: not enough values to unpack (expected 4, got 3)
```

This is because `red_cost_model.short_vectors` has a different number of outputs for `red_cost_model = Kyber` (3) vs 4 in all other cases.

bencrts:
This has been fixed (see #64 for comments).

PR #64, opened 2022-12-07
bugfix(lwe_dual): fix dual_hybrid for Kyber cost model

bencrts:
Fixes issue #63. 

Perhaps there is a cleaner way to do this, this is a quick fix.


hamishun (commented):
In this particular case, if the only difference is the number of trailing unpacked variables that you don't care about you then you can just use the '*' way. see suggestions, no extra logic required.

malb:
I *think* this has been resolved by @hamishun in the meantime? Can you check?

PR #65, opened 2022-12-21
refactor to simplify tasks in util.py

hamishun:
The code in `batch_estimate` is noisy due to the datatype conversions. 
This PR:
- simplifies the code around tasks and results
- removes some unnecessary for loops
- adds a named tuple for limits/bounds to make access to those types more intuitive
- define the `Pool` in a context manager to be properly cleaned up

### DIFF
--- a/estimator/lwe.py
+++ b/estimator/lwe.py
@@ -19,8 +19,8 @@ from .reduction import RC
 
 
 class Estimate:
-    @classmethod
-    def rough(cls, params, jobs=1, catch_exceptions=True):
+
+    def rough(self, params, jobs=1, catch_exceptions=True):
         """
         This function makes the following somewhat routine assumptions:
 
@@ -82,18 +82,19 @@ class Estimate:
         }
 
         for algorithm in algorithms:
-            for k, v in res.items():
-                if algorithm == k:
-                    if v["rop"] == oo:
-                        continue
-                    print(f"{algorithm:20s} :: {v!r}")
+            if algorithm not in res:
+                continue
+            result = res[algorithm]
+            if result["rop"] != oo:
+                print(f"{algorithm:20s} :: {result!r}")
+
         return res
 
     def __call__(
         self,
         params,
-        red_cost_model=None,
-        red_shape_model=None,
+        red_cost_model=red_cost_model_default,
+        red_shape_model=red_shape_model_default,
         deny_list=tuple(),
         add_list=tuple(),
         jobs=1,
@@ -124,11 +125,6 @@ class Estimate:
 
         """
         params = params.normalize()
-
-        if red_cost_model is None:
-            red_cost_model = red_cost_model_default
-        if red_shape_model is None:
-            red_shape_model = red_shape_model_default
 
         algorithms = {}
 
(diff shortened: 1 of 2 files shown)


estimator/lwe.py
```diff
@@ -4,18 +4,23 @@
...
 from .gb import arora_gb  # noqa
 from .lwe_parameters import LWEParameters as Parameters  # noqa
+from .conf import (red_cost_model as red_cost_model_default,
+                   red_shape_model as red_shape_model_default)
+from .util import batch_estimate, f_name
+from .reduction import RC
 
 
 class Estimate:
-    @classmethod
```
malb, line 17: Why?

estimator/lwe.py
```diff
@@ -42,12 +47,6 @@ def rough(cls, params, jobs=1, catch_exceptions=True):
 
 
         """
-        # NOTE: Don't import these at the top-level to avoid circular imports
```
malb, line 45: how did you resolve this?

estimator/util.py
```diff
@@ -366,28 +391,20 @@ def batch_estimate(params, algorithm, jobs=1, log_level=0, catch_exceptions=True
...
     else:
-        pool = Pool(jobs)
-        res = pool.starmap(_batch_estimatef, tasks)
-        res = {(f_repr, x): res[i] for i, (_, x, _, f_repr, _) in enumerate(tasks)}
-
-    ret = {x: {} for _, x in res.keys()}
-    for (f, x), v in res.items():
-        if v is not None:
-            ret[x][f] = v
+        with Pool(min(jobs, cpu_count())) as pool:
```
malb, line 407: Why put the limit there?

hamishun: Why have the limit or why specifically put it there and not elsewhere?
For the former, if you have more job processes than cpus that can run them, your throughput will go down. For the latter, it seems like the place to naturally put it; where the `Pool` is created. Thinking about it though, maybe a writting to the log in case that the number of processes are clipped to the max number may be useful.

estimator/lwe.py
hamishun: I can revert the change if you prefer. 

Too large an explanation for such a minor thing :D. My reasoning was that I don't see when this `rough` method is used from the class i.e. line 39 in the example doctest has `lwe.estimate.rough(schemes.Kyber512)` making use of the global `estimate` instance. This makes sense as the other method `__call__` is an instance method. Therefore `rough` is subtly not acting as most users would expect/interpret, it seems surprising. The `cls` first argument is not required by the implementation; therefore, an argument could be made that the method could be decorated as a staticmethod. However, again the situation that `__call__` would be an instance method with only one way to be called and `rough` could be called either from the class or an instance seems inconsistent.

estimator/lwe.py
hamishun: Always happy to revert the change.
I didn't observe any circular dependecies. `flake8` didn't pick up on anything (not 100% guarantee). I have a couple of thoughts on the matter, 
1. Importing in this way is contrary to PEP8. Clarity and least surprise is to aim to always import at the top of the file.
2. Circular dependency imports should not be ignored through workarounds. These workarounds ideally should be temporary measures; they are usually seen as code smells indicating a serious problem with the structure of the codebase. Instead of pre-empting, letting the problem show itself and addressing it is better in the long run for a project.

estimator/util.py
malb: Ah, sorry. I meant why impose a limit at all. I'd say if a user wants to use 200 jobs on a system that reports 8 threads, let them?

estimator/lwe.py
malb: Okay, let's see how it goes :)

estimator/lwe.py
malb: Yeah, okay, I see the lack of symmetry here. Let's do it as you suggest. Thanks

estimator/util.py
hamishun: Ok. Change reverted :)

malb:
Thank you!

PR #66, opened 2023-01-04
refactor Cost as specialised dictionary inheriting from UserDict

hamishun:
Instances of Cost are used as a specialised dictionary.
Changes:
- Cost inherits from UserDict simplifying the implementation instead of manipulating an object's underlying dict. Inheriting from UserDict instead of dict directly ensures correct behaviour from known issues.
- Removal of several '# noqa'
- Simplification of some other code.

### DIFF
--- a/estimator/lwe_bkw.py
+++ b/estimator/lwe_bkw.py
@@ -27,7 +27,7 @@ class CodedBKW:
 
     @staticmethod
     @cached_function
-    def ntest(n, ell, t1, t2, b, q):  # noqa
+    def ntest(n, ell, t1, t2, b, q):
         """
         If the parameter ``ntest`` is not provided, we use this function to estimate it.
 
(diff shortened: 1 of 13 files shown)


estimator/lwe_bkw.py
```diff
@@ -72,7 +72,7 @@ def t1(params: LWEParameters, ell, t2, b, ntest=None):
             ntest = CodedBKW.ntest(params.n, ell, t1, t2, b, params.q)
         sigma_set = sqrt(params.q ** (2 * (1 - ell / ntest)) / 12)
         Ni = [CodedBKW.N(i, sigma_set, b, params.q) for i in range(1, t2 + 1)]
-        t1 = len([e for e in Ni if e <= b])
+        t1 = sum(e <= b for e in Ni)
```
malb, line 75: This is a rare case of me thinking the old version was more readable.

malb:
Really nice, thanks! I won't push back if you insist on the one change flagged above.

estimator/lwe_bkw.py
hamishun: @malb , I understand what you mean. Personally, I think they are both not ideal :D. However, I would champion the change purely because it does not require creating another list. Bools are guaranteed to work as expected here as their instances inherit from numbers.Integral.

As a compromise I have resorted to adding a clarifying comment of what the line of code is doing; otherwise, quite happy to revert the change.

malb:
Thanks!

PR #72, opened 2023-03-13
Constant c fix

### LINKED ISSUE #71: Kyber reduction model seems to use the wrong constant C in quantum mode
The [definition](https://github.com/malb/lattice-estimator/blob/48fa49bac864f48b8efe9c1c3ace3f67d3dcc675/estimator/reduction.py#L726) of the constant C uses 2^{0.292 β + o(β)} as the cost of sieving. But in quantum mode, the time complexity of sieving becomes 2^{0.265 β + o(β)} and the constant C should become 1/(1 - 2^{-0.265}) = 5.95. This will result in an error of about 0.1 bit when estimating with quantum mode like the [example](https://github.com/malb/lattice-estimator/blob/48fa49bac864f48b8efe9c1c3ace3f67d3dcc675/estimator/reduction.py#L734):
`Kyber(nn="list_decoding-ge19")(500, 1024)`
  I think the constant C might not be needed as a parameter. We can easily compute it by `C = 1.0/(1.0 - 2**(-self.NN_AGPS[self.nn]["a"]))`

shal10w:
- convert the parameter C to a variable computed in the function
- recompute some examples

the estimated results in classical model seems to change too.Because the default value of a is `NN_AGPS["list_decoding-classical"]["A"]` = 0.296, not 0.292. 
fixes #71


malb:
Thanks and sorry for the delay, could you rebase and resubmit. I'll respond quicker this time, I promise!

malb:
Rebased and merged, thanks

PR #76, opened 2023-07-30
Fix slow Sage imports

grhkm21:
By replacing all Sage imports with an individual imoprt, the loading time for `from estimator import *` reduces from 3.5s to 1.4s. Doctests seem to work (except for timeout stuffs) so I assume I didn't break anything 😄

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,5 +38,5 @@ jobs:
           # sage wants it
           sudo mkdir -p /usr/share/sagemath/build/pkgs
           # run tests
-          PYTHONIOENCODING=UTF-8 PYTHONPATH=`pwd` python -m pytest
+          PYTHONIOENCODING=UTF-8 PYTHONPATH=`pwd` python3 -m pytest
           
--- a/estimator/cost.py
+++ b/estimator/cost.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from collections import UserDict
 
+from sage.rings.all import QQ  # noqa: F401
 from sage.functions.log import log
 from sage.misc.functional import round
 from sage.rings.infinity import infinity as oo
--- a/estimator/cost.py
+++ b/estimator/cost.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 from collections import UserDict
 
-from sage.all import log, oo, round
+from sage.functions.log import log
+from sage.misc.functional import round
+from sage.rings.infinity import infinity as oo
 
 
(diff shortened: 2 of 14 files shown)


bencrts:
I guess we would need the CI to pass before we used an alternative technique for importing (see here: https://github.com/malb/lattice-estimator/pull/76/checks). It seems to be unhappy with the way `infinity` is imported (I can't tell why at first look).

grhkm21:
The PEP8 fails because of `F401 'sage.rings.all.QQ' imported but unused`, but this is necessary due to Sage shenanigans. I just overwrote it with `# noqa`, which allows the check to be performed for other import statements.

Regarding the infinity import problem, I have no idea why. It works for me locally:
```python
sage -python -c "from sage.rings.infinity import infinity as oo" 
```

malb:
Perhaps the CI uses a different Sage version? Those deep imports are not supported by Sage, ie it gives no guarantee those will work.

grhkm21:
The most recent log prints this at the very start:
```

Run export SAGE_ROOT=`sage -c "import os; print(os.environ['SAGE_ROOT'])" | tail -1`
Traceback (most recent call last):
  File "/usr/bin/sage-eval", line 10, in <module>
    eval(compile(s,'<cmdline>','exec'))
  File "<cmdline>", line 1, in <module>
  File "/usr/lib/python3.10/os.py", line 680, in __getitem__
    raise KeyError(key) from None
KeyError: 'SAGE_ROOT'
```
which seems weird. Also, CI seems to specify that Sage is installed via `apt-get install`, which from my past experience severely lags behind. It says [on this website](https://pkgs.org/download/sagemath) that it can be anywhere from Sage 9.0 to 10.0...
Is it possible to make CI use the [sage docker](https://hub.docker.com/r/sagemath/sagemath) image instead, or would that be too much work? This PR isn't *that* important after all

malb:
Indeed, this is installing `sagemath (9.5-4)`. I'm hesitant to make changes that will break on Sage versions this recent, just imagine all the panicking support requests we'd have to deal with.

grhkm21:
This is very weird, I tried using a fresh Sage 9.0 (conda) installation and there's no issue with importing. Not sure then...

I also noticed that at the start of the log, it prints an error where `SAGE_ROOT` env variable is not set, not sure if that is related.

I will close this issue but if I figure out why this is happening I will report back

malb:
FWIW, removing the `SAGE_ROOT` stuff from the tests doesn't affect them: https://github.com/malb/lattice-estimator/pull/81

PR #89, opened 2023-10-30
Change memory usage of dual attack

### LINKED ISSUE #88: Memory usage of DualHybrid attack
The current estimation for the memory usage in the DualHybrid attack, is equal to the time of the [FFT solver :link: ](https://github.com/malb/lattice-estimator/blob/main/estimator/lwe_dual.py#L238). However, I have some questions regarding the time and memory costs.
1) Shouldn't the memory usage of the lattice sieve be taken into account? That is, [the number of dual vectors `N` :link:](https://github.com/malb/lattice-estimator/blob/main/estimator/lwe_dual.py#L165) should somehow be taken into account for the memory. For example, the below code example suggests a 1:1 ratio between runtime and memory, whereas the FFT table is much smaller than the runtime:

```sage
from estimator import *
from estimator.lwe_dual import DualHybrid
from estimator.cost import Cost

Cost.register_impermanent(rop=True,mem=False,red=True,beta=False,delta=False,m=True,d=False,zeta=False,t=False)
DualHybrid.cost(LWE.exhaustive_search, schemes.Kyber512, 400, 22, 0, 80, log_level=1)

```
Output: `rop: ≈2^148.2, mem: ≈2^148.1, m: 510, β: 400, t: 80, d: 1000, ↻: 1`.

2) By sequentially performing the FFT solver (with an in-place FFT) for `size` many enumeration targets, leads instead to a memory cost of `size_fft` instead of `size * (m + t * size_fft)`, if you assume a score entry in the FFT table is 1 unit of memory.
3) I currently, do not understand the origin of the time-cost `size * m` in the FFT solver. Using [[Sec.5.2, MATZOV'22] :link:](https://zenodo.org/records/6412487), the updating of the "initial" FFT table can be reduced to an amortized cost of `4N` additions, when iterating over all the `size` possible enumeration guesses. That would lead to a total runtime cost of `4 * N * size` for initializing all the tables, instead of the naïve `t * N * size`, where `t` is the number of secret coordinates to guess mod 2.

ludopulles:
Resolves #88 

- The parameter `use_lll` in `lwe_dual.py` was unused.
- The `short_vectors` methods in `reduction.py` now consistently return a 4-tuple, containing sieving dimension.
- The memory cost of dual attack considers memory for storing the dual vectors, **AND** the FFT table.
- The runtime cost also takes into account building the initial FFT table, based on all the enumeration targets.
- Various typo fixes.
- Added note that the number of dual vectors required is assuming the Independence Heuristic.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,7 +35,7 @@ Quick Start
     bdd                  :: rop: ≈2^140.3, red: ≈2^139.7, svp: ≈2^138.8, β: 391, η: 421, d: 1013, tag: bdd
     bdd_hybrid           :: rop: ≈2^140.3, red: ≈2^139.7, svp: ≈2^138.8, β: 391, η: 421, ζ: 0, |S|: 1, d: 1016, prob: 1, ↻: 1, tag: hybrid
     bdd_mitm_hybrid      :: rop: ≈2^260.3, red: ≈2^259.4, svp: ≈2^259.3, β: 405, η: 2, ζ: 102, |S|: ≈2^247.2, d: 923, prob: ≈2^-113.8, ↻: ≈2^116.0, tag: hybrid
-    dual                 :: rop: ≈2^149.9, mem: ≈2^88.0, m: 512, β: 424, d: 1024, ↻: 1, tag: dual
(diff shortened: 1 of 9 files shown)


blowfish880 (approved):
Looks good to me. Thanks! @malb I'd be happy with a merge if you are.

PR #90, opened 2023-11-07
Add NotImplementedError to primal_dsd for Xs != Xe

hkippen-SBAQ:
Add NotImplemented error for primal_DSD when distributions do not match. Commit also updates the API documentation to include ntru.rst, giving tutorial code for NTRU attack estimation.


malb:
Thanks!

Issue #110, opened 2024-05-07
dual crash


bencrts:
The param set:
```
sage: param = LWE.Parameters(
....:     n = 4096, 
....:     q = 2**125,
....:     # Xs = ND.SparseTernary(32768, 64, 64),
....:     Xs = ND.Uniform(-1, 1),
....:     Xe = ND.DiscreteGaussian(3.19),
....:     m = oo,
....:     tag = "param_1024_gaussian_classic_128"
....: )
```

crashes with `LWE.dual_hybrid(param)`, e.g.

```
OverflowError                             Traceback (most recent call last)
<ipython-input-6-70ea24a47b40> in <module>
----> 1 LWE.dual_hybrid(param)

~/Desktop/Github/lattice-estimator/estimator/lwe_dual.py in __call__(self, params, red_cost_model, log_level)
    654                     with local_minimum(40, params.n, log_level=log_level + 4) as it:
    655                         for beta in it:
--> 656                             cost = self.cost(
    657                                 beta,
    658                                 params,

~/Desktop/Github/lattice-estimator/estimator/lwe_dual.py in cost(cls, beta, params, m, p, k_enum, k_fft, beta_sieve, red_cost_model)
    592             p,
    593         )
--> 594         rho, T_sample, _, beta_sieve = red_cost_model.short_vectors(
    595             beta, N=N, d=k_lat + m, sieve_dim=beta_sieve
    596         )

~/Desktop/Github/lattice-estimator/estimator/reduction.py in short_vectors(self, beta, d, N, preprocess, B, C, sieve_dim)
    894
    895         c = N / floor(2 ** (0.2075 * sieve_dim))
--> 896         sieve_cost = C * 2 ** (self.NN_AGPS[self.nn]["a"] * sieve_dim + self.NN_AGPS[self.nn]["b"])
    897         return (
    898             rho,

OverflowError: (34, 'Result too large')
```

it's similar to a few previous bugs, I think we need some `RR()s`. I've self-assigned it.

PR #112, opened 2024-05-08
fix Issue 110

bencrts:
**Issue 1.**
As in #110, the crash was caused by some missing `RR()s` which have been added below. After this, the code outlined in the issue runs but segfaults. The segfault happens for both `RC.BDGL16` and `RC.MATZOV`, but for different reasons.

**Issue 2.**
After fixing 1. Variants of `short_vectors()` are crashing for different cost models and causing a segfault, this happens when `c` in the below code is huge (e.g. `>2**large`). Basically, we are calling e.g. `ceil(2**large)` and it just crashes. To avoid this, I picked an arbitrary value and set c to `oo` in this case. The choice of the arbitrary value (`2**1000`) might need some discussion.

I also added a blankline to `lwe_dual.py` for readability.

### DIFF
--- a/estimator/lwe_dual.py
+++ b/estimator/lwe_dual.py
@@ -591,6 +591,7 @@ class MATZOV:
             k_fft,
             p,
         )
+
         rho, T_sample, _, beta_sieve = red_cost_model.short_vectors(
             beta, N=N, d=k_lat + m, sieve_dim=beta_sieve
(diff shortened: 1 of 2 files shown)


bencrts:
Made a small update but I think it's OK now

malb:
Looks good to me.

PR #113, opened 2024-05-08
Another fix + add params to doctest

bencrts:
Another `RR()` fix which fixes the same error as #110 for a different parameter set / model combination. I've also added another HE-like parameter set to the doctests to help catch this stuff, it's medium-sized (`n=8192`) and doesn't add much time to the CI run.


bencrts:
Keeping in draft for now.

malb:
lgtm

PR #117, opened 2024-07-17
include ChaLoy21 = ChaLoy21() in Class RC in reduction.py

gong-cr:
Got error message: AttributeError: type object 'RC' has no attribute 'ChaLoy21' when selecting quantum cost model = RC.ChaLoy21.


malb:
Ta!

PR #122, opened 2024-09-12
include chaloy bkz time estimation in the form as LaaMosPol14

gong-cr:
Add BKZ total time estimation for ChaLoy21 as ChaLoy21_BKZ.
Two cost models:
ChaLoy21: time estimation for quantum sieving algorithm (single SVP call within BKZ) based on the work [AC:ChaLoy21].
ChaLoy21_BKZ: total time estimation for BKZ including multiple calls to SVP oracle using sieving algorithm from [AC:ChaLoy21].

### DIFF
--- a/estimator/reduction.py
+++ b/estimator/reduction.py
@@ -650,13 +650,12 @@ class ADPS16(ReductionCost):
 
 class ChaLoy21(ReductionCost):
 
-    __name__ = "ChaLoy21"
+    __name__ = "ChaLoy21_CoreSVP"
     short_vectors = ReductionCost._short_vectors_sieve
 
     def __call__(self, beta, d, B=None):
         """
-
-        See [AC:ChaLoy21]_.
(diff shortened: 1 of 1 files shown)


malb:
Where does this formula come from? I'd be surprised if `(0.2570 * beta + 16.4 + log(self.svp_repeat(beta, d), 2))` was correct.

gong-cr:
The BKZ cost cost follows the same structure as in BDGL16 (classical) and LaaMosPol14 (quantum), but the coefficient is adjusted to align with the time complexity of the ChaLoy21 quantum sieving algorithm.  I agree that the factor 16.4 may not be very solid, and other terms might also be questionable under the quantum regime. I’m not sure if there are any better alternatives at this point.

malb:
Mhhh, now that I think about it, I don't think we should cost Quantum BKZ, i.e. change `LaaMosPol14`.

Issue #95, opened 2024-01-25
RuntimeError: β = 1050 > d = 1049 for n=1030, q=2^64


ElenaKirshanova:
Executing

`params=LWE.Parameters(n=1030, q=2^64, Xs=ND.UniformMod(2), Xe=ND.DiscreteGaussian(stddev=3.19))
primal_usvp_cost = LWE.primal_usvp(params, red_cost_model=RC.BDGL16)`

does not pass sanity_check():

`RuntimeError: β = 1050 > d = 1049`

Oddly, for the default red_cost_model (red_cost_model = RC.MATZOV) the issue does not appear. 

Potential fix: checking inside _solve_for_d https://github.com/malb/lattice-estimator/blob/eae2fdb794eb5e1eb9dacd1ee13b62f6bdeb71b2/estimator/lwe_primal.py#L40 that the returned d is at least beta.

ElenaKirshanova:
Just noticed that it can be related to https://github.com/malb/lattice-estimator/issues/87#issue-1947852243

Issue #87, opened 2023-10-17
Estimation fail caused by β > d for small LWE


nindanaoto:
Dear the maintainer, 

I'm now trying to estimate the security for small LWE ciphertexts like the following:

```
param = estimator.lwe_parameters.LWEParameters(
    n=384,
    q=2 ** 7,
    Xs=estimator.nd.NoiseDistribution.UniformMod(2),
    Xe=estimator.nd.NoiseDistribution.CenteredBinomial(8),
    m = 384
)
r = estimator.LWE.estimate(param,red_cost_model = estimator.RC.BDGL16)
```
This code gives an exception caused by β > d for primal_uvsp and η > d. I know that they are checked by sanity_check in the cost.py. However, I think setting β and η to be β = d  and η = d is better than raising the exception. Is there any reason for avoiding such a fix? 

This code also raises exceptions for Incorrect bounds for primal_bdd but I'm not sure how to fix this.

ryanleh:
I had a similar issue (see #93) that was fixed by downgrading Sage to v9.5.

nindanaoto:
Really? I will try it. Thanks!

Issue #106, opened 2024-03-19
BDD Crash


bencrts:
BDD crash leads to no estimates (at all) being output, e.g.:

```
params = LWE.Parameters(n=700, q=2**64, Xs=ND.UniformMod(2), Xe=ND.UniformMod(2**64//2**5))
LWE.Estimate(params)
KeyError                                  Traceback (most recent call last)
<ipython-input-4-7507ca3b58d0> in <module>
----> 1 LWE.estimate(params)

~/Desktop/Github/lattice-estimator/estimator/lwe.py in __call__(self, params, red_cost_model, red_shape_model, deny_list, add_list, jobs, catch_exceptions)
    166             if result["rop"] == oo:
    167                 continue
--> 168             if algorithm == "bdd_hybrid" and res["bdd"]["rop"] <= result["rop"]:
    169                 continue
    170             if algorithm == "bdd_mitm_hybrid" and res["bdd_hybrid"]["rop"] <= result["rop"]:

KeyError: 'bdd'
```

usvp alone runs:

```
sage: LWE.primal_usvp(params)
rop: ≈2^251.6, red: ≈2^251.6, δ: 1.002423, β: 797, d: 1046, tag: usvp
```

I can see why it happens and will take a look at it, just opening an issue to track it.

bencrts:
This seems fine on current `main` now, closing!

PR #126, opened 2024-09-20
Potential fix for #87 and #95

ludopulles:
Two clamping changes to hopefully fix the two issues above. As it's a change in the important part `primal_usvp` of the lattice estimator, I would appreciate someone double-checking if I am not breaking stuff with these changes :smile: .

- Explanation `d = max(d, beta)`: `_solve_for_d` computes the optimal `d` using the ABC-formula (aka solving a quadratic equation), and makes sure this value `d` is in between `n` and `m`, but it does not consider `beta` here. By increasing `d`, you simply take a "suboptimal" `d` but one in which you can actually run BKZ-beta.

- Explanation `if eta > d`: In some cases it turns out `beta` is low, you have weak BKZ-beat lattice reduction, which does allow you to solve the BDD instance "even with a SVP-call in dimension `eta = d + 1`" (i.e. the two curves in [PKC:PosVir21]_ do not intersect, not even at non-integer points). Ergo, try bigger block sizes `beta` in `cost_zeta` (the level above `cost`). However, I'm not entirely sure this infinity-cost is handled properly in `cost_zeta`. The CI doctests suggest it does work.

Closes: #87
Closes: #95

### DIFF
--- a/estimator/lwe_primal.py
+++ b/estimator/lwe_primal.py
@@ -99,6 +99,8 @@ class PrimalUSVP:
             tau = False  # Tau false ==> instance is homogeneous
 
         d = PrimalUSVP._solve_for_d(params, m, beta, tau, xi) if d is None else d
+        if d < beta:
+            d = beta
         # if d == β we assume one SVP call, otherwise poly calls. This makes the cost curve jump, so
         # we avoid it here.
         if d == beta and d < m:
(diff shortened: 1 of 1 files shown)


ludopulles:
Also fixes #106:

```sage
from estimator import *
params = LWE.Parameters(n=700, q=2**64, Xs=ND.UniformMod(2), Xe=ND.UniformMod(2**64//2**5))
LWE.primal_bdd(params)
```
Output: `rop: ≈2^263.2, red: ≈2^42.8, svp: ≈2^263.2, β: 40, η: 867, d: 867, tag: bdd`

Closes: #106

malb:
This looks good to me, could we add some tests to show that this fixes these corner cases? Just to prevent future regressions

ludopulles:
Of course, they are added!

malb:
Looks good to me. @bencrts do you agree?

bencrts:
Yep! Thanks @ludopulles

PR #127, opened 2024-09-20
Thoroughly change NoiseDistribution

ludopulles:
This pull request contains a lot of modifications to the code base, in particular the file `estimator/nd.py` has been changed. The main contribution is four-fold:

1. Currently, `NoiseDistribution` is one bulky class containing all the code specifics for different distributions: `DiscreteGaussian, Uniform, CenteredBinomial and SparseTernary` and has in various places case-by-case logic to return the correct value.
This pull-request makes NoiseDistribution look like an "interface", for which there are many instantations, and helper functions (`DiscreteGaussianAlpha, UniformMod, SparseBinary, Binary, Ternary`) which call these instances. This improves usability of ND's, extendability to more distributions/functions, and flexibility in attacks that target specific ND's (i.e. like `MITM` in `estimator/lwe_guess.py` that has code for `SparseTernary`).

2. Moreover, there was a `TODO` in the code of SparseTernary that access to the variables `p` and `m`  was needed to compute the actual `support_size`. That is now very easy, so I did that. The code in `estimator/lwe_guess.py` has been adjusted properly to compute the right probability, but also by moving this logic to `SparseTernary`, such that it may well be used in different places (`lwe_dual.py` perhaps?). This changes the attack costs, by generally lowering it by 1 or 2 bits of security (the search space is a `~6√3/(πn)`-fraction smaller), but also weirdly increases sometimes (probably: when doing MitM, you split the secret in two parts, one of the two might have a higher density -> bigger search space).

3. The parameter `n` does not need to be given to a NoiseDistribution (especially a secret!) when creating a LWEParameter, because `__post_init__` sets `n` to the right value. Thus, remove `n` in all example codes. Especially, in the docs the secret was called with a *different* `n` than the `n` of LWEParameters. This shows that this redundancy should be removed to prevent errors. Note: I also moved `n` to last parameter in `SparseTernary` to make this possible, and to make it more consistent with the other constructors.

4. Small fixes in the documentation.

Random quirk: running the doctest on estimator/util.py gave me an error. I wonder how this passed CI.


estimator/lwe_dual.py
```diff
@@ -377,13 +373,13 @@ def __call__(
             >>> dual_hybrid(params, mitm_optimization="numerical")
             rop: ≈2^129.0, m: 1145, k: 1, mem: ≈2^131.0, ↻: 1, β: 346, d: 2044, ζ: 125, tag: dual_mitm_hybrid
 
-            >>> params = params.updated(Xs=ND.SparseTernary(params.n, 32))
+            >>> params = params.updated(Xs=ND.SparseTernary(32))
             >>> LWE.dual(params)
             rop: ≈2^103.4, mem: ≈2^63.9, m: 904, β: 251, d: 1928, ↻: 1, tag: dual
             >>> dual_hybrid(params)
-            rop: ≈2^92.1, mem: ≈2^78.2, m: 716, β: 170, d: 1464, ↻: 1989, ζ: 276, h1: 8, tag: dual_hybrid
+            rop: ≈2^91.6, mem: ≈2^77.2, m: 711, β: 168, d: 1456, ↻: ≈2^11.2, ζ: 279, h1: 8, tag: dual_hybrid
```
malb, line 380: Why does this change?

estimator/lwe_dual.py
```diff
@@ -428,7 +424,7 @@ def __call__(
 
         params = params.normalize()
 
-        if params.Xs.is_sparse:
+        if type(params.Xs) is SparseTernary:
```
malb, line 427: `is_sparse` could cover more than merely `SparseTernary`

ludopulles: With the current sparsity cutoff at 1/2, I think the only other sparse noise distributions are:
- `Uniform(x, x)`: a constant value is not a secret so this makes LWE trivially broken, and the code will produce some errors.
- DiscreteGaussian with a very small `stddev`: I believe the current code does not accurately handle such Discrete Gaussians. Namely, experimenting with stddev=0.5 show me that `density` and `support_size` are currently computing under- and overestimates. (I'm happy to create an issue for this if needed!)

The reason I changed it to SparseTernary is to 1) take the number of +1's and -1's into account and 2) remove the calculation using the bounds as these are giving overestimates on the actual support size for the above case of the DiscreteGaussian.

Let me split up this PR into only refactoring ND, and I'll put the SparseTernary changes into a separate one.

estimator/lwe_dual.py
ludopulles: This is because the search space of sparse ternaries becomes smaller when you take the number of +1's and -1's into account. Related to below, I'll also put these changes to the docs into a different PR.

estimator/lwe_dual.py
malb: Sounds good. FWIW I was also thinking about future proving the code :)

ludopulles: Hum, I created a branch for just refactoring ND: 
https://github.com/malb/lattice-estimator/compare/main...ludopulles:lattice-estimator:refactor-ND

But it seems there are some weird cases, where SparseTernary is called with `p` a half-integer. Is it okay to not split up the PR, but instead to keep the generic "is_sparse" code with a special case for SparseTernary inside there?

malb: What do you mean by: "keep the generic "is_sparse" code with a special case for SparseTernary inside there" ? An attribute that checks "isinstance(SparseTernary)"?

ludopulles: Basically this commit: 31957df62dddc4517e1a54f46a3812741c12398d

I meant to revert the code back to "if params.Xs.is_sparse:", and then inside this if-statement add code specific for SparseTernary to arrive at the right estimates, because the previous generic sparse secret code did not take into account the weight and +1/-1 conditions for SparseTernary.

bencrts:
Just a note that #130 will have some merge conflicts with this, I can always wait for this PR to be merged and rebase over there based on any changes.

ludopulles:
Thanks for the heads up, it should be fairly easy to make it into a class after this PR.

I think I'm done here. I'm waiting to know if the changes in here are OK:
- [estimator/lwe_dual.py:L66-L73](https://github.com/malb/lattice-estimator/pull/127/files#diff-c5afdd8d23bcfc4d102dd30888b41c1779cea8f4765c180b10c71ea28ac6c401R66-R73)
- [estimator/lwe_guess.py:L243-L252](https://github.com/malb/lattice-estimator/pull/127/files#diff-6f8d120656947bb038ba8f6e920c565193ef4ce7ebdbe42d02d74d39c39a8c7bR243-R252)
- [estimator/lwe_guess.py:L290-L300](https://github.com/malb/lattice-estimator/pull/127/files#diff-6f8d120656947bb038ba8f6e920c565193ef4ce7ebdbe42d02d74d39c39a8c7bR290-R300)

malb:
I'm happy to merge this, unless there's something on your todo list before?

malb:
It's really nice, thank you!

ludopulles:
There are no TODOs for now, so ready to merge :smiley:.

PR #130, opened 2024-09-25
Add TUniform distribution

bencrts:
Adding the TUniform distribution which is used in the tfhe-rs library (as well as other FHE schemes, such as BGV/CKKS). This is essentially a generalisation (consider `b=0`) of the distribution which samples:

```
-1 with prob 1/4
1 with prob 1/4
0 with prob 1/2
```

which is commonly mixed up with the uniform ternary distribution. More generically, the distribution `TUniform(-2**b, 2**b)` samples:

```
-2**b, 2**b each with probability 1/2**(b+2)
the other terms with probability 1/2**(b+1)
```

For small values of `b`, the associated standard deviation is different enough from the `Uniform()` distribution that it is worth including both.

### DIFF
--- a/estimator/nd.py
+++ b/estimator/nd.py
@@ -380,6 +380,58 @@ class Uniform(NoiseDistribution):
         a, b = self.bounds
         return ceil(RR(fraction) * (b - a + 1)**len(self))
 
+class TUniform(NoiseDistribution):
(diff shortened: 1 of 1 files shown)


malb:
lgtm

PR #131, opened 2024-10-16
Fix MitM Babai e-admissability computation

ludopulles:
- The e^(-x^2) forgot to take the square in the exponent,
- there was a unnecessary change from stddev to standard width 'alphaq',
- negative values don't occur because the formula is right :),
- applying some scaling to (r_i, \sigma) such that \sigma = 1/\sqrt{2} makes formulas easier to parse.


malb:
I'm fixing the doctest failures on main.

estimator/prob.py
```diff
@@ -78,35 +77,26 @@ def gaussian_cdf(mu, sigma, t):
...
-        RR(
-            erf(s * sqrt(RR(pi)) / alphaq)
-            + (alphaq / s) * ((exp(-s * sqrt(RR(pi)) / alphaq) - 1) / RR(pi))
-        )
-        for s in map(sqrt, r)
-    )
-    p = RR(prod(probs))
-    return p if 0 <= p <= 1 else 0.0
+    # Note: `r` contains *square norms*, so convert to non-square norms.
+    # Follow the proof of Lemma 4.2 [EPRINT_SonChe19]_, because that one uses standard deviation.
```
malb, line 95: Nit: should read [EPRINT:SonChe19] ?

malb:
Also: thank you!

estimator/prob.py
ludopulles: Good catch. While I was at it, I changed the abbreviation EPRINT -> IEEE (as well as 3 other refs).

malb:
Ta!

bencrts:
Ooh. This is a nice catch @ludopulles!

Issue #111, opened 2024-05-08
why only enforce infinity norm < q rather than (q-1)/2?


josephjohnston:
For SIS, infinity norm beta implies solution values can lie in the region [-beta,beta] and therefore with negative values I assume we're using signed modulus range [-(q-1)/2,(q-1)/2]. So if beta is not enforced to be less than (q-1)/2 then solutions should always be trivial since they have access to the full modulus range. Yet the estimator only asks that beta < q (and can give large rop/rep for beta barely below q, eg q-2). Since it yields high rop/rep estimates for trivial instance params, I'm not sure how to judge what it yields for instance params I'd imagine are non-trivial.
Why this behavior? Any explanation would be great.

josephjohnston:
Here's an example.
```python
beta=2^16-1
SIS.estimate(SIS.Parameters(n=2^10,q=2^16+1,length_bound=beta,norm=Infinity))
# yields: lattice  :: rop: ≈2^134.0, red: ≈2^133.0, sieve: ≈2^133.0, β: 364, η: 372, ζ: 488, d: 1440, prob: 1, ↻: 1, tag: infinity
```
Since this is reports non-trivial time for a trivial instance I'm not sure how small beta must be to actually make the instance non-trivial.

malb:
Agreed, this looks like a bug. I put it on the TODO list.

psivesely:
Closed in #139.

PR #139, opened 2025-01-27
Correct trivial SIS cases in sis_lattice.py

psivesely:
A(q*e1) = 0 mod q is a trivial solution for when the input bound is >= q for both l_2 and l_inf norms. In the centered representation we're using signed modulus range [-(q-1)/2,(q-1)/2] as clarified in https://github.com/malb/lattice-estimator/issues/111


malb:
Thanks. Maybe we should distinguish between ISIS and SIS here?

PR #177, opened 2025-11-12
Remove usage of 'is' operator for boolean comparison

cyrilbouvier:
The ['is' operator](https://docs.python.org/3/reference/expressions.html#is-not) is used in some `if` statements in the code to test equalities with boolean values (`condition is True`, `condition is not True`, ...). This is too restrictive and can lead to weird behaviors (see example below).
This PR proposes to replace all occurrences of `if condition is True`, `if condition is False`, `if condition is not True`, `if condition is not False` by the more general [boolean statements](https://docs.python.org/3/reference/expressions.html#boolean-operations) `if condition` and `if not condition`, which perform boolean conversion contrary to the `is` operator.

## Example

Consider the following code:

```py
from estimator import *

RingDegree = 256
MatrixDim = 3
SumSize = 2
SumNumber = 2**5
Modulus=11017217
MaskingLevel = 8
variance = (2**SumSize)*sqrt((SumNumber-MaskingLevel+1)/6) #20*sqrt(1/6)
n = RingDegree * MatrixDim


for var in (variance, float(variance)):
  print(f"#### {var=}")
  nd = ND.NoiseDistribution(None, 0, var)
  params = LWE.Parameters(n=n, q=Modulus, Xs=nd, Xe=nd,m=n)
  LWE.estimate(params)
```

In the first iterator the variance is a symbolic expression from SageMath, in the second iteration the variance is the corresponding float value. The two estimates return very different values for uvsp: $2^{43.0}$ versus $2^{141.4}$.
The difference is only due to the [condition line 997 of `reduction.py`](https://github.com/malb/lattice-estimator/blob/352ddaf4a288a0543f5d9eb588d2f89c7acec463/estimator/reduction.py#L997):
```py
if predicate is False:
  ...
```
The function `cost` is called by `cost_gsa` with `predicate = lhs <= rhs`. By using the `is` operator, conversion to boolean are prevented. So when using symbolic expression from SageMath, `predicate` is not a boolean but a symbolic expression which is convertible to a boolean with a meaningful truth value but the conversion is prevented by the usage of the `is` operator. By replacing the condition by
```py
if not predicate:
  ...
```
fixed the problem.

### DIFF
--- a/estimator/cost.py
+++ b/estimator/cost.py
@@ -85,14 +85,14 @@ class Cost(UserDict):
                     vv = "%7s" % ("≈2^%.1f" % log(v, 2))
             except TypeError:  # strings and such
                 vv = "%8s" % v
-            if compact is True:
+            if compact:
                 kk = kk.strip()
                 vv = vv.strip()
             return f"{kk}: {vv}"
 
         # we store the problem instance in a cost object for reference
         s = [value_str(k, v) for k, v in self.items() if k != "problem"]
-        delimiter = "\n" if newline is True else ", "
+        delimiter = "\n" if newline else ", "
         return delimiter.join(s)
 
     def reorder(self, *args):
--- a/estimator/cost.py
+++ b/estimator/cost.py
@@ -190,7 +190,7 @@ class Cost(UserDict):
         return Cost(**cost)
 
     def __bool__(self):
-        return self.get("rop", oo) < oo
+        return bool(self.get("rop", oo) < oo)
 
     def __add__(self, other):
         return self.combine(self, other)
--- a/estimator/lwe_dual.py
+++ b/estimator/lwe_dual.py
@@ -289,7 +289,7 @@ class DualHybrid:
             log_level=log_level,
         )
 
-        if fft is True:
+        if fft:
 
             def f(beta):
                 with local_minimum(0, params.n - zeta) as it:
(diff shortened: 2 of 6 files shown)


cyrilbouvier:
CI reveals another bug that was hidden by the use of `if res is not False and ...` in https://github.com/malb/lattice-estimator/blob/352ddaf4a288a0543f5d9eb588d2f89c7acec463/estimator/util.py#L180 

Consider the following code (shorter code from a .ipynb tested by CI):
```py
from estimator import *
from estimator.lwe_dual import matzov
params = LWE.Parameters(n=200, q=7981, Xs=ND.SparseTernary(16), Xe=ND.CenteredBinomial(4))
matzov(params)
```
I do not know why but the cost computed by this function is of type `sage.symbolic.expression.Expression` (but holds the non symbolic value `7.55864263177626e25`). So when using
```py
if res and ...
```
instead of
```
if res is not False and ...
```
in the `update` method of the class `local_minimum_base` in `utils.py`, it raises an exception because the `__bool__` method of the `Cost` class return `self.get("rop", oo) < oo` which, when the left hand side is of type `Expression`, is not a boolean (as required by Python) but another `Expression`.
Using `if res is not False ...` would always be true in this case (because res is never the boolean `False`, it is of type `Cost`). In fact, the `__bool__` method is never called in this case as there is no conversion to boolean with the `is` operator.

Note: the same arguments apply to the `update` method of `early_abort_range` in `utils.py`

Commita a2be406 fixes this bug.

TabOg:
Heya, unfortunately we merged this but then had to revert. Would you be able to open a new PR from this branch, ideally after updating to main?

PR #193, opened 2026-03-01
Adding Large_Norm (DucEspPos23)

navid-divan:
This implements the small modulus (I)SIS method from "Finding Short Integer Solutions When the Modulus Is Small" (Ducas, Espitau, Postlethwaite, CRYPTO 2023). The attack uses Z-shape geometry, which exists in BKZ-reduced q-ary lattices when their length bound ν reaches or exceeds q. The method performs q-vector analysis through projection because it uses the projected sublattice to obtain short vectors that exist in the remaining q-vectors. The method applies when ν reaches or exceeds q/2.


malb:
Sorry it took me so long to take a first look! But it seems this isn't ready for inclusion as it re-implements a bunch of stuff that is already in the estimator, i.e. this PR doesn't use the estimator machinery.

navid-divan:
> Sorry it took me so long to take a first look! But it seems this isn't ready for inclusion as it re-implements a bunch of stuff that is already in the estimator, i.e. this PR doesn't use the estimator machinery.

Probably because I stuck too closely to the proof of concept in https://github.com/verdiverdiverdi/ISIS-small-q, I'll update it by doing it from scratch soon.

docs/algorithms/sis-lattice.rst
```diff
@@ -33,4 +33,19 @@ Another option is to simulate a rerandomization of the basis, such that the q-ve
     SIS.lattice(params.updated(length_bound=70), red_shape_model=Simulator.LGSA)
 
 **Note:** Currently, lattice attack estimation is only available for euclidean (``2``) and infinity (``oo``) norms. ``SIS.lattice()`` will return a ``NotImplementedError`` if one of these two norms are not selected.
-                        
+
+Small Modulus
```
malb: I disagree with the authors of [C:DucEspPos23] that the key magnitude here is the size of $q$ rather than the norm bound $\nu$. 

@verdiverdiverdi can chip in and set me straight.

docs/algorithms/sis-lattice.rst
```diff
@@ -33,4 +33,19 @@ Another option is to simulate a rerandomization of the basis, such that the q-ve
...
+-------------
+
+When the modulus ``q`` is small relative to a euclidean length bound ``ν > q``, plain SIS is trivially solved by a q-vector, but finding a solution that is not a multiple of ``q`` (the SIS\* problem) remains hard. The attack of [C:DucEspPos23]_ exploits the Z-shape of a BKZ-reduced basis of the q-ary lattice: it sieves in the projected sublattice past the q-vectors and lifts the resulting short vectors over them, which succeeds when the (uniform mod q) lifted entries are short enough. The number of such short lifts is counted via theta series of ``ZZ``::
+
+    from estimator import *
+    params = SIS.Parameters(n=150, q=257, length_bound=420, m=300, norm=2)
+    SIS.small_q(params)
+
+The reduction of [C:DucEspPos23]_ frames signature forgery for schemes such as Falcon and Mitaka as an inhomogeneous instance; pass ``inhom="specific"`` (or ``inhom="generic"``) to account for the corresponding loss in success probability::
```
malb: I'd say, we should not cost reductions here, but attacks. If we want to cost ISIS, we should have a new class for ISIS.

estimator/sis_small_q.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+    Squares of a centred set of representatives of ``ZZ/qZZ``.
+
+    These are the squared lengths a single (uniform mod q) lifted coordinate can take, used to form
+    the cumulant generating function to the squared length of a lift.
+
+    :param q: the modulus.
+
+    """
+    u = np.arange(-((q - 1) // 2), q // 2 + 1, dtype=float)
```
malb: This code is linear in $q$, which I'd say is too expensive (think super-poly $q$).

estimator/sis_small_q.py
```diff
@@ -0,0 +1,302 @@
...
+        Cost of the small-q attack [C:DucEspPos23] for a fixed block size β.
+
+        The attack reduces a basis of the kernel lattice ``Λ_q^⊥(A)``, which exhibits a "Z-shape":
+        a head of ``n_q`` vectors of length ``q``, a sloped middle and a flat tail. Sieving in the
+        projected sublattice past the q-vectors yields many short vectors which are then lifted over
+        the q-vectors; a lift is a solution when its (uniform mod q) lifted entries are short enough.
+
+        :param beta: block size β.
+        :param params: SIS parameters.
+        :param on_the_fly: lift every vector seen in the last sieve iteration [C:DucEspPos23]_ Sect. 4.
```
malb: Do we need the `on_the_fly` parameter?

estimator/sis_small_q.py
```diff
@@ -0,0 +1,302 @@
...
+            proj_length = sqrt(RR(4) / 3) * params.q
+            log2_density = log2(RR(4) / 3) / 2
+
+        if proj_length >= params.length_bound:
+            # the projected vectors alone are already too long to lift to a solution
+            return Cost(rop=oo)
+
+        # shape of the BKZ-β reduced basis of the kernel lattice Λ_q^⊥(A)
+        simulator = simulator_normalize(red_shape_model)
+        if simulator is GSA:
```
malb: I'm not convinced this hard-coded behaviour is great. If there's a silent switch, it should happen at a higher level.

docs/algorithms/sis-lattice.rst
navid-divan: @malb, so, if I have understood correctly, you are saying that the binding quantity is the norm bound relative to the modulus (the attack is selected precisely on ν > q). Then, I will update it on condition ν > q instead of the absolute size of q.

docs/algorithms/sis-lattice.rst
navid-divan: OK, I will remove the `inhom` parameter together with its ISIS probability-loss factors and the `small_q_isis` tag and modify it to cost attacks instead of a problem-to-problem reduction into a cost.

estimator/sis_small_q.py
navid-divan: Hmmm… I think the sum over all q residues should be replaced with a closed-form coordinate modelled as uniform on $`[-q/2, q/2]`$, which is $`O(1)`$ in $`q`$. I will do it.

estimator/sis_small_q.py
navid-divan: Well, maybe it is unnecessary, since the estimator should output the strongest attack, and I am going to remove the parameter to force the attack to always take the cheaper of the two lifting.

estimator/sis_small_q.py
navid-divan: I see. I will remove the in-attack silent switch and make `SIS.estimate` pass the Z-shape simulator at registration.

docs/algorithms/sis-lattice.rst
malb: You're already checking for the norm bound relative to q, no?

navid-divan: Yes, it has been `length_bound > q​`, so there was no big change to make; my only edit to this was rephrasing the doc so it opens with "ν exceeds q" instead of "q is small" to align my wording with the condition that the code already gives. I just wanted to be precise :)

estimator/sis_small_q.py
```diff
@@ -0,0 +1,253 @@
...
+def log2_lift_proportion(n_q, sq_radius, q):
+    """
+    Base-2 logarithm of the proportion of ``Cube_{n_q}(q)`` lying inside a Euclidean ball.
+
+    ``Cube_{n_q}(q)`` is a centred set of representatives of ``(ZZ/qZZ)^{n_q}`` and the proportion of
+    it inside the ball of squared radius ``sq_radius`` is the success probability of a single lift
+    [C:DucEspPos23]_ Eq. 2, whose integer points [C:DucEspPos23]_ Sect. 3.3 counts via theta series.
+    We instead use a saddle-point (Bahadur-Rao) approximation: a coordinate uniform mod q is modelled
+    as uniform on ``[-q/2, q/2]``, whose cumulant generating function ``K(t) = log E[e^{t U^2}]`` is
+    closed form, so the estimate is accurate in the relevant tail and costs ``O(1)`` rather than
```
malb: I think there should be at least also a slow function that computes this quantity directly instead of approximating it and there should be tests asserting that the two functions give values that are close. As it stands, it's a lot of work to convince yourself that this function is at least approximately correct.

estimator/sis_small_q.py
```diff
@@ -0,0 +1,253 @@
+# -*- coding: utf-8 -*-
+"""
+Estimate cost of solving SIS with a small modulus via lattice reduction [C:DucEspPos23]_.
+
+See :ref:`SIS Lattice Attacks` for an introduction to what is available.
+
+"""
+from functools import partial
+from math import log as mlog, sqrt as msqrt, exp as mexp, pi, erf
```
malb, line 9: Why? Shouldn't we use Sage's multiprecision tools instead?

estimator/sis_small_q.py
```diff
@@ -0,0 +1,253 @@
...
+
+        if beta > d:
+            return Cost(rop=oo)
+
+        # shape of the BKZ-β reduced basis of the kernel lattice Λ_q^⊥(A)
+        simulator = simulator_normalize(red_shape_model)
+        r = simulator(d=d, n=d - params.n, q=params.q, beta=beta, xi=1, tau=False)
+
+        # number of q-vectors n_q left at the head of the basis (Zone I of the Z-shape)
+        if abs(sqrt(r[0]) - params.q) < 1e-6:  # q-vectors exist
```
malb: This is the second ad hoc constant, perhaps there should be one for precision and that should be a variable at the top of the file?

estimator/sis_small_q.py
navid-divan: Done. I added `log2_lift_proportion_exact` to count the points directly via theta-series convolution.

estimator/sis_small_q.py
navid-divan: OK, in the updated version, I have considered the q-vector–detection tolerance to be a named module constant `profile_precision` at the top of the file.

estimator/sis_small_q.py
navid-divan: @malb, I kept `math/scipy` intentionally: Sage's `RR` defaults to the same 53-bit (double) precision, so there's no accuracy gain for what is ultimately a cost reported in bits, and when I tried Sage's `find_root` for the internal root-finding it failed to locate the root on valid inputs (the root lies extremely close to zero, in a narrow region Sage's solver scans past) whereas `scipy.brentq` finds it reliably. So, I guess switching would only reduce robustness, and `math/scipy` are already used elsewhere in the estimator (e.g., `simulator.py`, `reduction.py`) as well.

malb: RR would allow us to up the precision if needed. Do we know we won't need it?

estimator/sis_small_q.py
```diff
@@ -0,0 +1,316 @@
...
+        # the ball reaches past the mean cube length, so the lifting is not the bottleneck
+        return RR(0)
+    if a < 1:
+        # only the zero vector of the cube lies in the ball
+        return RR(-n_q * log2(q))
+
+    def derivatives(t):
+        # K'(t), K''(t) of K(t) = log E[e^{t U^2}] via I_k = ∫_0^h u^k e^{t u^2} du for one coordinate
+        s = -t
+        if s < 1e-8:
```
malb: same here, should this be the same precision as above or a different variable?

estimator/sis_small_q.py
```diff
@@ -0,0 +1,316 @@
...
+        probability = RR(2) ** min(RR(0), log2_solutions)
+
+        # cost of producing the short vectors and the underlying BKZ-β reduction
+        rho, cost_red, N, sieve_dim = red_cost_model.short_vectors(beta, d, sieve_dim=sieve_dim)
+        bkz_cost = costf(red_cost_model, beta, d)
+
+        ret = Cost()
+        ret["rop"] = cost_red
+        ret["red"] = bkz_cost["rop"]
+        ret["sieve"] = max(cost_red - bkz_cost["rop"], 1e-100)  # ensuring non-zero cost here
```
malb, line 237: Just say 1?

estimator/sis_small_q.py
navid-divan: @malb, yes, we know. The saddle point's own model error is larger than the double-precision arithmetic error. So extra float precision cannot make this function any more accurate; it would just be computing an approximate model to more decimals. The only lever for genuine accuracy is the exact count, which I have already provided. The way I see it, higher precision is unobservable here.

estimator/sis_small_q.py
navid-divan: It should be different, and they must not be shared. `profile_precision` is a tolerance for comparing lengths when detecting q-vectors in the basis profile; `1e-8` is the cutoff on the saddle parameter below which the closed-form integrals degenerate to 0/0. I will update it now and give it its own named constant `small_s_cutoff` to make that clear.

estimator/sis_small_q.py
navid-divan: @malb, I took `max(..., 1e-100)` exactly (value and comment) from the `sis_lattice.py`, where the floor makes a negligible sieve-cost difference print as ≈0 (2^-332) rather than a one-operation cost, so I think it might be better to keep the two SIS cost functions consistent and change both together if at all, rather than diverging here.

estimator/sis.py
```diff
@@ -48,9 +49,12 @@ def rough(self, params, jobs=1, catch_exceptions=True, quiet=False):
         """
         algorithms = {}
 
-        # Only lattice attacks are supported on SIS for now
         algorithms["lattice"] = partial(lattice, red_cost_model=RC.ADPS16, red_shape_model="lgsa")
 
+        # the small modulus attack [C:DucEspPos23]_ applies to euclidean instances with ν > q
```
verdiverdiverdi: Perhaps missing a comment regarding the zgsa here, as below.

estimator/sis_small_q.py
```diff
@@ -0,0 +1,321 @@
...
+    """
+    Exact base-2 logarithm of the proportion of ``Cube_{n_q}(q)`` lying inside a Euclidean ball.
+
+    This counts the integer points directly, by convolving truncated theta series of ``ZZ`` as in
+    [C:DucEspPos23]_ Sect. 3.3. It is a slow reference, of cost ``O(n_q · sq_radius · q)``, against
+    which :func:`log2_lift_proportion` is validated; the latter should be preferred in practice.
+
+    :param n_q: number of (uniform mod q) lifted coordinates.
+    :param sq_radius: squared radius of the ball.
+    :param q: the modulus.
```
verdiverdiverdi, line 42: Does the code below handle the even q case, which has a 'scare' footnote on p.15 of https://eprint.iacr.org/2023/1125.pdf?

estimator/sis_small_q.py
```diff
@@ -0,0 +1,321 @@
...
+
+    """
+    import numpy as np
+
+    if sq_radius < 0:
+        return RR(-oo)
+    half = (q - 1) // 2
+    max_sq = n_q * half ** 2
+    D = min(int(floor(sq_radius)), max_sq)
+    if D >= max_sq:
```
verdiverdiverdi: I find this a little obscured. Why not

if sq_radius >= max_sq:
    ...

and only define D below, e.g. after l. 58?
In the cases where the ball does entirely contain the box, D will equal max_sq, so the >= comparison is also a little confusing.

estimator/sis_small_q.py
```diff
@@ -0,0 +1,321 @@
...
+        # the whole cube lies inside the ball, every lift is short enough
+        return RR(0)
+
+    # squared-length distribution of one centred coordinate uniform mod q, truncated beyond X^D
+    dist = np.zeros(D + 1)
+    for u in range(-half, q // 2 + 1):
```
verdiverdiverdi: Why not half + 1 for the upper limit for the range?
Does this (silently) capture the even q case?

estimator/sis_small_q.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+def log2_lift_proportion(n_q, sq_radius, q):
+    """
+    Base-2 logarithm of the proportion of ``Cube_{n_q}(q)`` lying inside a Euclidean ball.
+
+    ``Cube_{n_q}(q)`` is a centred set of representatives of ``(ZZ/qZZ)^{n_q}`` and the proportion of
+    it inside the ball of squared radius ``sq_radius`` is the success probability of a single lift
+    [C:DucEspPos23]_ Eq. 2. Rather than counting integer points directly as in
+    :func:`log2_lift_proportion_exact`, we use a saddle-point (Bahadur-Rao) approximation: a
```
verdiverdiverdi: This is a novel technique to me, thank you for introducing it to me.
Thinking about the geometry of the problem, in a single dimension a reduced point lives in [-(q-1)/2, (q-1)/2] intersect ZZ (for odd q).
To approximate this by a (continuous) uniform random variable on some interval, it seems reasonable to give each integer point above a tile of ZZ, e.g. (-1/2, 1/2].
Is this how you ended up with [-q/2,q/2]?

estimator/sis_small_q.py
```diff
@@ -0,0 +1,321 @@
...
+    coordinate uniform mod q is modelled as uniform on ``[-q/2, q/2]``, whose cumulant generating
+    function ``K(t) = log E[e^{t U^2}]`` is closed form, so the estimate costs ``O(1)`` rather than
```
verdiverdiverdi: U is undefined.

estimator/sis_small_q.py
```diff
@@ -0,0 +1,321 @@
...
+        >>> approx = log2_lift_proportion(60, 40000, 127)
+        >>> exact = log2_lift_proportion_exact(60, 40000, 127)
+        >>> bool(abs(approx - exact) < 0.1)
+        True
+
+    """
+    a = float(sq_radius)
+    h = q / 2.0  # a lifted coordinate is modelled as uniform on [-q/2, q/2], with E[U^2] = h^2 / 3
+    if a >= n_q * h ** 2 / 3:
+        # the ball reaches past the mean cube length, so the lifting is not the bottleneck
```
verdiverdiverdi: This intuitively seems reasonable, but slightly non trivial to prove.
Have you, or has it been empirically stress tested?

estimator/sis_small_q.py
verdiverdiverdi: (is there a reference to exactly the version you implement below?)

estimator/sis_small_q.py
```diff
@@ -0,0 +1,321 @@
...
+        )
+        # repeat the whole attack ~1/prob times, rerandomising the basis between attempts
+        if probability and not RR(probability).is_NaN():
+            ret = ret.repeat(prob_amplify(success_probability, probability))
+        else:
+            return Cost(rop=oo)
+
+        return ret
+
+    def __call__(
```
verdiverdiverdi, line 259: As a sanity check, does your code produce similar (the same?) numbers as Tab. 2 in https://eprint.iacr.org/2023/1125.pdf, specifically for SIS* with and without otf?

estimator/sis.py
navid-divan: OK, I will also expand the comment in `rough()` to state that `small_q` gets the zgsa simulator.

estimator/sis_small_q.py
navid-divan: I just checked the footnote on p. 15. The representative range in my code already produced the non-symmetric even-`q` cube `{-(q-2)/2,…,q/2} `of Def. 16. But I will fix the bound `max_sq` to use `(q//2)²` rather than `((q-1)//2)²`; though, as the footnote 3 says, all moduli of interest here are odd anyway.

estimator/sis_small_q.py
navid-divan: I did not have a special reason for this. So, I should early-return on `sq_radius >= max_sq` directly and define `D` only afterwards to remove the confusion on `D >= max_sq` comparison; correct?

estimator/sis_small_q.py
navid-divan: Maybe I should add a comment about this. Yes, `q//2 + 1` upper limit captures even `q` because, for odd `q`, it coincides with half+1, but for even `q`, it extends the range to include the point `q/2`, which gives the non-symmetric centred cube. I will add a comment to clarify it.

estimator/sis_small_q.py
navid-divan: > Is this how you ended up with [-q/2,q/2]?

Yes, exactly!

> is there a reference to exactly the version you implement below?

Yes and no; this technique seems classical to me (I will add https://doi.org/10.1214/aoms/1177705674 to references), but its application to this specific count is validated against the exact function rather than taken from a code reference.

estimator/sis_small_q.py
navid-divan: Will do ASAP.

estimator/sis_small_q.py
navid-divan: Yes, I did; in the lower tail, it agrees with `<0.05 bit`, and the `a ≥ mean` shortcut has at most `~1` bit of error right at the mean (it returns `p=1` where the true value is `~½`), which is conservative. As you mentioned, below is my stress test for q=257, n_q=120, mean=660480.
a/mean | approx | exact | note
-- | -- | -- | --
0.30 | 2^−78.15 | 2^−78.16 | tail: <0.01 bit
0.60 | 2^−23.37 | 2^−23.41 | tail: ~0.04 bit
0.90 | 2^−2.73 | 2^−3.19 | ~0.46 bit
1.00 | 2^0 (shortcut) | 2^−0.99 | ~1 bit, conservative
1.10 | 2^0 (shortcut) | 2^−0.17 | shrinking
1.50 | 2^0 | 2^−0.0 | exact

And for the reproduction of SIS*, CoreSVP/ADPS16 to compare with Table 2 of [DEP23](https://link.springer.com/chapter/10.1007/978-3-031-38548-3_6) I get:

scheme | no-otf (mine/paper) | otf (mine/paper)
-- | -- | --
Falcon q=257 | 94.6 / 95 | 89.9 / 90
Falcon q=521 | 112.1 / 113 | 109.2 / 110
Falcon q=1031 | 116.8 / 117 | 119.1 / 120
Mitaka q=521 | 54.3 / 55 | 50.8 / 51
Mitaka q=1031 | 80.6 / 81 | 77.7 / 78

estimator/sis_small_q.py
navid-divan: @verdiverdiverdi, Yes, as I have mentioned above, it reproduces them for SIS* within ~1 bit in every case, and with the public `SIS.small_q`, it reports the cheaper of the two as the attack cost.

estimator/sis_small_q.py
verdiverdiverdi: Yes please.

verdiverdiverdi:
I do not have strong feelings about the name, though I would say big beta captures the situation no better or worse than small q.
In particular, the efficacy of the attack is really a function of the ratio between modulus and length bound, i.e. 'small q [relative to beta]' or 'big beta [relative to q]' is required.
(Also, just checking that beta is not reserved for blocksize?)

We called the paper '...when the modulus is small', in the abstract have '...for anomalously small moduli q...' and called the original implementation [1] 'ISIS-small-q', so anyone who has interacted with those materials may expect some type of small q nomenclature.
However, introduction in the lattice-estimator will certainly set the naming convention going forwards, so it can be whatever you both think is best.

[1]: https://github.com/verdiverdiverdi/ISIS-small-q

PR #137, opened 2024-12-18
perf(prob): improve performance of hybrid bdd

bencrts:
Use `erf, sqrt, exp` from `numpy` for faster computation in the mitm probability for the hybrid-decoding attack. 

We see an improvement for `primal_hybrid` from 113 seconds down to 25 seconds, which is about ~4.5x. Similar improvements are seen for the other parameter sets that I have tried. 

Leaving some performance stats below for future reference.

Test code:

```
from estimator import *

params = LWE.Parameters(n=4096, q = 2**64, Xs = ND.CenteredBinomial(1), Xe = ND.CenteredBinomial(1))
from time import time
start = time()
est = LWE.primal_hybrid(params)
end = time()
print(end - start)
print(est)

```

New stats

```
25.026100873947144
         rop: ≈2^309.0
         red: ≈2^308.1
         svp: ≈2^307.8
           β:      653
           η:        2
           ζ:      280
         |S|: ≈2^380.1
           d:     7773
        prob: ≈2^-89.7
           ↻: ≈2^91.9
         tag:   hybrid
         16101142 function calls (16062506 primitive calls) in 27.260 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      241    6.943    0.029    7.393    0.031 simulator.py:112(<listcomp>)
      221    4.770    0.022    6.058    0.027 prob.py:117(<listcomp>)
      241    4.249    0.018    4.677    0.019 simulator.py:111(<listcomp>)
      345    2.432    0.007    2.432    0.007 {built-in method numpy.array}
  5858932    1.223    0.000    1.758    0.000 <frozen importlib._bootstrap>:389(parent)
    15675    0.832    0.000    0.832    0.000 other.py:756(_do_sqrt)
      451    0.818    0.002    0.820    0.002 {built-in method _imp.create_dynamic}
    15221    0.641    0.000    1.083    0.000 prob.py:121(drop)
  5872433    0.538    0.000    0.538    0.000 {method 'rpartition' of 'str' objects}
  1405891    0.438    0.000    0.438    0.000 {method 'cum_distribution_fu
[...]
```

Current main branch stats

```
113.00009608268738
         rop: ≈2^309.0
         red: ≈2^308.1
         svp: ≈2^307.8
           β:      653
           η:        2
           ζ:      280
         |S|: ≈2^380.1
           d:     7773
        prob: ≈2^-89.7
           ↻: ≈2^91.9
         tag:   hybrid
[...]

PR #198, opened 2026-04-11
Improve performance of the primal hybrid attack

bencrts:
Two things here:

1) Similar to #137, but used in a slightly more stable situation. We use numpy in the GSA computation, but since the values here are in the log domain there is no precision issue. 

2) I'm a bit less confident about this one, but after doing some profiling it looks like string formatting in logging is a bit of a performance issue, as can be seen below with `value_str` taking up to 20% of the running time. As best as I can tell (I haven't really interacted with the logger before), some strings were being formatted even when the logging level is not enabled. I think I narrowed it down to two places in utils.py and changed them so the formatting only happens when the logging level requires it.

Running `LWE.primal_hybrid(params, mitm=True, babai=False)` on these parameter sets:
```
params_1 = LWE.Parameters(n=2048, q=2**54,  Xs=ND.SparseTernary(128, 128), Xe=ND.DiscreteGaussian(3.19))
params_2 = LWE.Parameters(n=4096, q=2**109, Xs=ND.SparseTernary(128, 128), Xe=ND.DiscreteGaussian(3.19))
params_3 = LWE.Parameters(n=8192, q=2**218, Xs=ND.SparseTernary(128, 128), Xe=ND.DiscreteGaussian(3.19))
```
sees some nice improvements from these small changes. Here's the initial profile:

```
============================================================
  params_1
============================================================
         rop: ≈2^119.0
         red: ≈2^118.7
         svp: ≈2^116.4
           β:      288
           η:       18
           ζ:      299
         |S|: ≈2^175.1
           d:     3524
        prob:    0.118
           ↻:       37
         tag:   hybrid
         65272218 function calls (65181108 primitive calls) in 62.290 seconds

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  2241685   12.063    0.000   20.463    0.000 /./estimator/lwe_primal.py:377(svp_gaussian_heuristic_gsa)
   227303   11.265    0.000   12.073    0.000 /./estimator/cost.py:75(value_str)
      721    6.697    0.009    7.165    0.010 /./estimator/simulator.py:118(<listcomp>)
      721    5.356    0.007   26.532    0.037 /./estimator/lwe_primal.py:358(svp_dimension_gsa)
  ...



============================================================
  params_2
============================================================
         rop: ≈2^117.6
         red: ≈2^117.4
         svp: ≈2^114.5
           β:      286
           η:       36
           ζ:      418
         |S|: ≈2^158.0
           d:     7339
        prob:    0.375
           ↻:       10
         tag:   hybrid
         36420782 function calls (36371304 primitive calls) in 36.535 seconds

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      392    7.529    0.019    8.026    0.020 /./estimator/simulator.py:118(<listcomp>)
   133838    6.113    0.000    6.549    0.000 /./estimator/cost.py:75(value_str)
      392    5.245    0.013    6.652    0.017 /./estimator/prob.py:126(<listcomp>)
      392    4.535    0.012    5.013    0.013 /./estimator/simulator.py:117(<listcomp>)
  ...



============================================================
  params_3
============================================================
         rop: ≈2^118.2
         red: ≈2^118.0
         svp: ≈2^114.8
           β:      284
           η:       31
           ζ:      745
         |S|: ≈2^161.5
           d:    14845
        prob:    0.349
           ↻:       11
         tag:   hybrid
         52872855 function calls (52822084 primitive calls) in 51.776 seconds

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      347   13.315    0.038   14.208    0.041 /./estimator/simulator.py:118(<listcomp>)
      347    9.511    0.027   12.057    0.035 /./estimator/prob.py:126(<listcomp>)
      347    8.412    0.024    9.275    0.027 /./estimator/simulator.py:117(<listcomp>)
   143278    6.549    0.000    7.113    0.000 /./estimator/cost.py:75(value_str)
  ...
```

and these timings lower (without any changes to the estimates) from `(62.290s, 36.535s, 51.776s)` down to `(31.742s, 10.644s, 9.988s)` so up to a 5x gain. 

If 2) seems correct on review, I guess we can propagate this throughout the library for further performance gains.


TabOg:
Oh nice!

I am still seeing some overflow with large parameters like

```
params = LWEParameters(n=2 ** 15, q = 2 ** 767, Xs=ND.SparseTernary(p=96, m=96, n=2 ** 17), Xe=ND.DiscreteGaussian(stddev=3.19))
```

These are I believe used in OpenFHE. 

I get a numpy overflow in the simulator while calculating `4.0 ** r_arr`, because these numbers are bigger than the float. However the estimator still ran, because numpy warns rather than throws. Additionally the final estimate was still correct which is nice

I tried using numpy for the the part of the profile that was less than float precision, and sage for the rest, and then merging the two:

```
def GSA(d, n, q, beta, xi=1, tau=1, dual=False):
    from .reduction import delta as deltaf

    assert 2 <= beta <= d

    if not tau:
        log_vol = RR(log(q, 2) * (d - n) + log(xi, 2) * n)
    else:
        log_vol = RR(log(q, 2) * (d - n - 1) + log(xi, 2) * n + log(tau, 2))

    delta = deltaf(beta)
    log_delta = RR(log(delta, 2))
    # at this value, 4 ** r will overflow numpy float64
    numpy_threshold = np.finfo(np.float64).maxexp / 2
    # find value of i for which 4 ** r[i] stops overflowing
    threshold = floor((d - 1 - (numpy_threshold - log_vol/d) / log_delta) / 2) + 1
    first_safe_index = max(0, min(threshold, d))
    r_log_head = [(d - 1 - 2 * i) * log_delta + log_vol / d for i in range(first_safe_index)]
    r_head = [2 ** (2 * r_) for r_ in r_log_head]
    r_log_tail = (d - 1 - 2 * np.arange(first_safe_index, d)) * float(log_delta) + float(log_vol) / d
    r_tail = (4.0 ** r_log_tail).tolist()

    return r_head + r_tail
```
This still gives some speedup, but not as much: e.g. for those parameters, I go from 300 seconds on main -> 30 seconds with only numpy but overflows -> 110 seconds with this version. Also its quite hard to read :smiley:

bencrts:
Nice catch, I missed that. Yeah I agree that we definitely don't want overflows.

My guess is that the estimates are the same because (a) the mitm probability is set to 1 in this case (instead of being computed from the GSA output) and (b) the SVP dimension isn't impacted by the overflowing part of the basis. I guess that won't always be the case for all four of the primal-hybrid variants. 

I think the speed-up is worth slightly less readability: could be useful for those FHE tests. So, the change looks good to me. I can update the code with your suggestion: thanks!

PR #199, opened 2026-04-17
(schemes): add TFHE-rs parameters from v1.6

bencrts:
Adding TFHE-rs parameters to `schemes.py`. These are current parameters used in v1.6, and can be found here: 

https://github.com/zama-ai/tfhe-rs/blob/main/tfhe/src/shortint/parameters/v1_4/classic/tuniform/p_fail_2_minus_128/ks_pbs.rs#L29


bencrts:
closed (replaced by #208)

PR #200, opened 2026-04-21
using GSA inside bottleneck probabilities rather than in profile simulation

TabOg:
heya, I thought more about speeding up primal hybrid estimation with numpy.

I'm not so sure about using f64s for [a part of] the profile, as its not clear where those values will be used, and they might overflow there.

Instead I looked at using np inside the two bottleneck probabilities `prob_babai` and `mitm_babai_probability`, which are the only place the profile is actually used. I wrote two functions `prob_babai_gsa` and `mitm_babai_probability_gsa`.

Turns out, `prob_babai` can be calculated in f64s regardless of the input profile, as only the short vectors in the tail contribute to this probability.

`mitm_babai_probability` is harder -- I split the computation into two parts,  in the tail using np/scipy/f64s, and for the larger starting elements, using Sage again (but with some smaller speedups). 

While working on this I found that I'd actually made a mistake in the `prob_babai` function: I was using `(d - eta)` as the beta distribution parameter, when it should always be `d`. Have fixed this now...

With these changes, for the exact parameters you mentioned and babai=False/mitm=True, on my machine I get
`param_1: 50.39s, param_2: 16.96s, param_3: 12.89s`, whereas with just using numpy for profile, I get `param_1: 51.78s, param_2: 16.66s, param_3: 16.0s`, so no massive change. This might just be my machine, if you want to check its similar in your environment?

Looking instead at 
```
LWEParameters(n=2 ** 15, q = 2 ** 767, Xs=ND.SparseTernary(p=96, m=96, n=2 ** 17), Xe=ND.DiscreteGaussian(stddev=3.19))
```

On `main` I get:

```
babai=True, mitm=True: Time taken: 366.6542766094208 seconds
babai=True, mitm=False: Time taken: 204.4723732471466 seconds
babai=False, mitm=True: Time taken: 234.87016820907593 seconds
babai=False, mitm=False: Time taken: 299.9083902835846 seconds
```
whereas with these changes I get
```
babai=True, mitm=True: Time taken: 128.69082283973694 seconds
babai=True, mitm=False: Time taken: 2.0150210857391357 seconds
babai=False, mitm=True: Time taken: 14.450679779052734 seconds
babai=False, mitm=False: Time taken: 14.615266561508179 seconds
```
And produces the same estimates, so a massive speedup in some places, like 20x. 

`mitm=True` is still a huge bottleneck, if you have any ideas I can look into it further.

Let me know what you think!


TabOg:
Oh no the test I added on `babai_gsa` passes on my machine but fails on the machine where the tests run

```
    EXAMPLES::
        >>> from estimator import prob, schemes, simulator
        >>> from estimator.nd import SparseTernary
        >>> from estimator.reduction import delta as deltaf

        # check gsa and non-gsa probabilities match
        >>> params = schemes.Kyber512
        >>> beta = 300
        >>> r = simulator.GSA(d=2 * params.n, n=params.n,q=params.q, beta=beta, xi=1, tau=False, dual=True)

        >>> babai_dim = len(r)
        >>> log(prob.babai(babai_dim, r, sqrt(babai_dim) * params.Xe.stddev), 2)
        -330.9488591088554
        >>> log_delta = log(deltaf(beta))
        >>> short_vector_norm = sqrt(babai_dim) * params.Xe.stddev
        >>> log(prob.babai_gsa(babai_dim, short_vector_norm, 2 * params.n, 0.5 * log(prod(r)), log_delta), 2)
        -330.948859108856 # vs. -330.948859108932 
```

estimator/prob.py
```diff
@@ -111,20 +142,112 @@ def mitm_babai_probability(r, stddev, fast=False):
...
+    def prob(i):
+        return erf(exp(b - a * i)) - (1 - exp(-exp(2 * (b - a * i)))) / (exp(b - a * i) * sqrt_pi)
+    prob_large_xs = prod(prob(i) for i in range(threshold_i))
+
+    p = RR(prob_small_xs) * prob_large_xs
+    assert 0.0 <= p <= 1.0
+    return p
+
+
+def babai(babai_dim, r, norm):
```
malb, line 197: Can you add documentation for the parameters?

estimator/prob.py
```diff
@@ -111,20 +142,112 @@ def mitm_babai_probability(r, stddev, fast=False):
...
+        0.999951336768017
     """
     denom = float(2 * norm) ** 2
     T = RealDistribution("beta", ((len(r) - 1) / 2, 1.0 / 2))
-    probs = [1 - T.cum_distribution_function(1 - r_ / denom) for r_ in r]
-    return prod(probs)
+    return prod(1 - T.cum_distribution_function(1 - r_ / denom) for r_ in r[:babai_dim])
+
+
+def babai_gsa(babai_dim, norm, d, log_vol, log_delta):
```
malb, line 241: Same here, can you add documentation for the parameters?

estimator/prob.py
```diff
@@ -111,20 +142,112 @@ def mitm_babai_probability(r, stddev, fast=False):
...
+    """
+    # calculate log (r_i ** 2/ (2 * short_vector_len ** 2)) for each i up to babai_dim, assuming the GSA.
+    # We cast logs to floats, which shouldn't overflow.
+    a = 4 * log_delta
+    b = 2 * log_delta * (d - 1) + 2 * log_vol / d - 2 * log(2 * norm)
+    log_ratios = float(b) - float(a) * np.arange(babai_dim, dtype=np.float64)
+
+    # if some log_ratio >= 0, then the corresponding term in the product is 1, so we can ignore it.
+    log_ratios = log_ratios[log_ratios < 0]
+    ratios = np.exp(log_ratios)  # we know these won't overflow, because log_ratios < 0
```
malb, line 261: it could underflow, technically?

TabOg: Yeah its true. My hope was that if log_ratio is very negative, so that exp(log_ratio) is basically zero, it was already making the product essentially zero. Unlike the mitm probability, the babai probability is usually quite close to 1 for optimal beta

FWIW calculating this way was producing the correct probability for some sample parameters where the correct babai probability is 2^-330 

Do you happen to know what threshold I should compare to to catch underflows for f64s?

malb: I think that's fine, I don't think it's realistic that we carefully track underflows everywhere when they're not to be expected.

PR #201, opened 2026-04-28
Revert "Improve performance of the primal hybrid attack"

malb:
Reverts malb/lattice-estimator#198

image


malb:
https://github.com/malb/lattice-estimator/actions/runs/25045860284

PR #217, opened 2026-06-26
fix(sis): preserve tiny infinity probabilities

quangvdao:
## Motivation

I am developing a lattice-based PCS at https://github.com/LayerZero-Labs/akita. The scheme has an offline planning mode, wherein the planner would sweep through parameters, see if they are 128-bit secure according to SIS estimator in this repo, and then emit generated schedules which ends up in the preprocessing (and both prover & verifier takes them for the protocol structure).

Before this, we were setting SIS security based on L2 norm, through a L_inf => L2 norm conversion. This led to some fixes in this PR #213 that affect some edge cases we ran into. We are now moving to directly estimating in the infinity norm instead, which gives tighter bounds. Hence I opened the PRs (with AI assistance) that you closed; apologies for getting off on the wrong start.

## Particular changes

Now in this PR I have included two changes to fix the SIS estimator, in the infinity norm, in the edge cases where floating point behavior could lead to degenerate behavior (rounding to zero instead of returning a tiny number, leading to estimates printing `Infinity` due to division by zero).

The fixes here are:
- Computing `log(1-p)` via calling the library function `log1p` which computes `log(1+p)` and bakes in good edge case behavior. Otherwise, we could get `1 - p = 1` for tiny probabilities `p`
- Computing `1 - 2 * gaussian_cdf(0, sigma, -B)` directly via `erf(B / (sqrt(2) * sigma))`, which again leads to accurately returning a tiny number, instead of `gaussian_cdf(0, sigma, -B)` being too close to `1/2` when `B/sigma` is tiny

I have included an explicit SIS instance where these changes matter; this is in the doctest.

## Review suggestions

The changes I made are done by AI (but this PR description is written by me). If you disagree with the way the fixes are implemented, I'm happy to revise until the code quality is acceptable. I'd also appreciate any feedback about whether these fixes are sufficient or whether there is a better way to perform these computations in the tiny-result regime


estimator/prob.py
```diff
@@ -162,11 +180,26 @@ def amplify(target_success_probability, success_probability, majority=False):
...
+        >>> from estimator import prob
+        >>> from sage.all import floor, log, RR
+        >>> floor(log(prob.amplify(0.99, RR(2)**-10000), 2))
+        10002
     """
     if target_success_probability < success_probability:
         return ZZ(1)
     if success_probability == 0.0:
         return oo
+    if not majority:
```
malb, line 206: If the check is moved here, it should be removed below.

estimator/prob.py
```diff
@@ -151,6 +151,24 @@ def drop(n, h, k, fail=0, rotations=False):
...
+
+    TESTS::
+
+        >>> from estimator import prob
+        >>> from sage.all import floor, log
+        >>> floor(log(-1 / prob._log1mexp2(-10000), 2))
+        10000
+    """
+    if log_x >= 0:
+        return -oo
```
malb, line 172: shouldn't this raise an error for positive values?

quangvdao: yes it should, addressed

estimator/prob.py
quangvdao: thanks, I have removed below

estimator/prob.py
```diff
@@ -162,11 +186,26 @@ def amplify(target_success_probability, success_probability, majority=False):
...
+    if not majority:
+        try:
+            return ceil(
+                log(1 - target_success_probability)
+                / _log1mexp2(log(success_probability, 2))
+            )
+        except ValueError:
+            return oo
 
     prec = max(
```
TabOg: The current code sets a precision 'prec' that should mean we're safe to calculate `log(1-p_target) / log(1 - p)` with this precision: it looks like this change ignores this value? What value does `prec` take in the edge cases that are failing?

IMO, if we get rid of the manual precision tuning (and rely on the default precision of RR) in the case that majority = False, we should also get rid of it in the case majority = True

estimator/sis_lattice.py
```diff
@@ -131,6 +130,20 @@ def cost_infinity(
...
+            >>> from estimator.sis_lattice import SISLattice
+            >>> from estimator.sis_parameters import SISParameters
+            >>> from estimator.reduction import RC
+            >>> from estimator.simulator import LGSA
+            >>> from sage.all import oo
+            >>> params = SISParameters(n=32, q=2**220, m=128, length_bound=1, norm=oo)
+            >>> cost = SISLattice.cost_infinity(
+            ...     40, params, red_cost_model=RC.BDGL16, red_shape_model=LGSA, d=128
+            ... )
+            >>> cost["rop"] < oo and cost["prob"] > 0
```
TabOg: could this test be replaced with the actual value of cost? in the style of (e.g.) [here](https://github.com/malb/lattice-estimator/blob/27a581bb8e9d49f5e9e2db315bd48ac769d5f5f5/estimator/sis.py#L44) ?

estimator/sis_lattice.py
```diff
@@ -178,7 +193,7 @@ def cost_infinity(
             sigma = vector_length / sqrt(gaussian_coords)
 
             log_trial_prob = RR(
-                log(1 - 2 * gaussian_cdf(0, sigma, -params.length_bound), 2) * (gaussian_coords)
+                log(erf(params.length_bound / (sqrt(2) * sigma)), 2) * gaussian_coords
```
TabOg: ditto above comment about `gaussian_cdf`

estimator/prob.py
```diff
@@ -162,11 +186,26 @@ def amplify(target_success_probability, success_probability, majority=False):
...
+        >>> floor(log(prob.amplify(0.99, RR(2)**-10000), 2))
+        10002
     """
     if target_success_probability < success_probability:
         return ZZ(1)
     if success_probability == 0.0:
         return oo
+    if not majority:
+        try:
+            return ceil(
```
TabOg, line 209: hello! could we keep the comment explaining this calculation, 
```
# target_success_probability = 1 - (1-success_probability)^trials
```

estimator/sis_lattice.py
```diff
@@ -156,7 +169,9 @@ def cost_infinity(
             vector_length = rho * sqrt(r[0])
             # Find probability that all coordinates meet norm bound
             sigma = vector_length / sqrt(d_)
-            log_trial_prob = RR(d_ * log(1 - 2 * gaussian_cdf(0, sigma, -params.length_bound), 2))
+            log_trial_prob = RR(
```
TabOg: looking at the defn of `gaussian_cdf`, I think this could alternatively be fixed by changing `gaussian_cdf` to not cast to `RR`, and that way we could keep the function description for `gaussian_cdf`

estimator/sis_lattice.py
quangvdao: Addressed. The SIS regression now checks the complete expected `Cost` output rather than merely asserting that the cost and probability are finite. In particular, it records approximately `2^6994.0` operations and per-attempt probability `2^-6953.7`.

estimator/sis_lattice.py
quangvdao: Addressed in both probability branches. The SIS estimator continues to call `gaussian_cdf`; the numerical correction is now centralized in that helper rather than duplicating direct `erf` expressions at both call sites.

estimator/prob.py
quangvdao: Yes, this has been re-added

estimator/sis_lattice.py
quangvdao: Implemented—thank you for the suggestion. `gaussian_cdf` no longer wraps the expression in `RR`.

Keeping it symbolic allows

`1 - 2 * gaussian_cdf(0, sigma, -B)`

to simplify to the equivalent centered `erf` expression before numerical coercion, avoiding cancellation while preserving the existing abstraction and documentation.

I added a regression with `sigma = 2^100` and `B = 1` that verifies the resulting interval probability is positive and numerically equals `erf(B / (sqrt(2) * sigma))`.

estimator/prob.py
quangvdao: Agree & addressed. I have removed the manual RealField(prec) tuning from both branches.

The revision avoids forming 1 - p entirely: _log1mexp2(log2(p)) evaluates it as log1p(-p). This removes the cancellation without requiring precision proportional to -log2(p). The ordinary branch computes log(1-p) this way, while the majority branch computes log(1-p²) by passing 2 * log2(p) to the same helper. This preserves ordinary RR-level numerical accuracy; it is intended to eliminate the catastrophic zero/infinity failure.

More rationale: in the failing SIS instance, the per-attempt probability is approximately \(2^{-6953.718}\). The old heuristic computes

```
2 * ceil(|log2(p)|) = 13,908 bits
```

but then caps this with min(prec, 2048), so the precision actually used is only 2,048 bits. At that precision, p itself remains representable, but 1 - p rounds to exactly 1, making log(1 - p) = 0.

Similarly, for the explicit p = 2^-10000 regression, the heuristic requests 20,000 bits but again actually uses 2,048.

estimator/sis_lattice.py
quangvdao: UPDATE: this is no longer correct; see comment below

estimator/sis_lattice.py
quangvdao: UPDATE: this is not sufficient; see comment below

quangvdao:
UPDATE: removing the outer `RR(...)` coercion from `gaussian_cdf` is not enough.

In particular, the hope is that the caller’s expression

    1 - 2 * gaussian_cdf(0, sigma, -B)

will simplify symbolically to

    erf(B / (sqrt(2) * sigma))

before numerical evaluation. Further testing showed that this is not reliable: whether it works depends on the symbolic shape of `sigma`.

In the original regression, we used:

    sigma = RR(2) ** 100

Sage represented the error-function argument as:

    -(3.94430452610506e-31) * sqrt(2)

and left

    erf(-(3.94430452610506e-31) * sqrt(2))

unevaluated. Removing the outer coercion therefore allowed the surrounding expression to cancel algebraically before conversion to `RR`. That made the test pass.

The actual SIS instance has a different expression shape:

    sigma = 1.33252774905124e16 * sqrt(2)

Here, the `sqrt(2)` in the Gaussian CDF denominator cancels the `sqrt(2)` in `sigma`. Sage consequently evaluates the inner expression numerically:

    erf argument = -3.75226707553369e-17
    erf value    = -4.23397999741062e-17

It then evaluates the CDF construction as:

    1 + erf_value = 1.00000000000000
    CDF           = 0.500000000000000
    1 - 2 * CDF   = 0

Thus, the information is already lost during `1 + erf_value`, before the removed outer `RR(...)` coercion would have been applied. Although Sage reports these values as symbolic `Expression` objects, the `erf` call has already been replaced by an approximate numerical constant. Removing the final coercion therefore does not guarantee symbolic cancellation.

We could force the expression to remain symbolic with `erf(..., hold=True)` and exact rational coefficients. I do not think that is the best contract for `gaussian_cdf`, however: it changes a numeric helper into a symbolic one, introduces symbolic-expression overhead, and still makes correctness depend on Sage’s later simplification behavior.

I have instead restored `gaussian_cdf`’s original numeric behavior and added a dedicated helper:

    gaussian_centered_interval_probability(sigma, bound)

It evaluates the centered Gaussian mass directly as:

    RR(erf(bound / (sqrt(2) * sigma)))

This is algebraically equal to `1 - 2 * CDF(-bound)`, but it never forms either `1 + tiny_value` or `1 - nearly_one`, so there is no catastrophic cancellation. Both infinity-norm SIS paths now use this helper.

